Issue #47, opened 2023-09-02
Allow disabling of policy event logs


mycrEEpy:
The collection of policy event logs is pretty expensive due to being a busy loop (using unix.EpollWait).

Users which are not interested in the logs (and/or are not willing to use a full CPU core on each node for the busy loop) should have the possibility to disable the policy event logs.

Furthermore I suggest the default behavior should be to disable the policy event logs due to the high cost factor.

jdn5126:
Closing now that VPC CNI v1.15.1 is released on GitHub: https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.15.1. This release contains the Network Policy agent v1.0.4 image tag, which includes #48

PR #48, opened 2023-09-02
feat: Add flag enable-policy-event-logs

mycrEEpy:
*Issue #, if available:* #47

*Description of changes:*
Add flag `enable-policy-event-logs` to disable event logging due to high cost of one full CPU core on each node.
Policy event logging is also now disabled by default.

### DIFF
--- a/pkg/config/controller_config.go
+++ b/pkg/config/controller_config.go
@@ -9,6 +9,7 @@ const (
 	defaultLogLevel                = "info"
 	defaultLogFile                 = "/var/log/aws-routed-eni/network-policy-agent.log"
 	defaultMaxConcurrentReconciles = 3
+	flagEnablePolicyEventLogs      = "enable-policy-event-logs"
 	flagEnableCloudWatchLogs       = "enable-cloudwatch-logs"
 	flagEnableIPv6                 = "enable-ipv6"
 	flagEnableNetworkPolicy        = "enable-network-policy"
@@ -22,6 +23,8 @@ type ControllerConfig struct {
 	LogFile string
 	// MaxConcurrentReconciles specifies the max number of reconcile loops
 	MaxConcurrentReconciles int
+	// Enable Policy decision logs
+	EnablePolicyEventLogs bool
 	// Enable Policy decision logs streaming to CloudWatch
 	EnableCloudWatchLogs bool
 	// Enable IPv6 mode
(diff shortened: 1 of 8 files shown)


pkg/ebpf/events/events.go
```diff
@@ -175,48 +175,46 @@ func capturePolicyEvents(ringbufferdata <-chan []byte, log logr.Logger, enableCl
...
-					}
-					protocol := getProtocol(int(rb.Protocol))
-					verdict := getVerdict(int(rb.Verdict))
-
-					log.Info("Flow Info:  ", "Src IP", utils.ConvByteArrayToIP(rb.SourceIP), "Src Port", rb.SourcePort,
-						"Dest IP", utils.ConvByteArrayToIP(rb.DestIP), "Dest Port", rb.DestPort,
-						"Proto", protocol, "Verdict", verdict)
-
-					message = "Node: " + nodeName + ";" + "SIP: " + utils.ConvByteArrayToIP(rb.SourceIP) + ";" + "SPORT: " + strconv.Itoa(int(rb.SourcePort)) + ";" + "DIP: " + utils.ConvByteArrayToIP(rb.DestIP) + ";" + "DPORT: " + strconv.Itoa(int(rb.DestPort)) + ";" + "PROTOCOL: " + protocol + ";" + "PolicyVerdict: " + verdict
+		for record := range ringbufferdata {
```
mycrEEpy, line 178: Also changed this line to wait for data from the channel instead of dead spinning the CPU in the for loop.

pkg/config/controller_config.go
```diff
@@ -39,7 +42,8 @@ func (cfg *ControllerConfig) BindFlags(fs *pflag.FlagSet) {
 		"Set the controller log file - if not specified logs are written to stdout")
 	fs.IntVar(&cfg.MaxConcurrentReconciles, flagMaxConcurrentReconciles, defaultMaxConcurrentReconciles, ""+
 		"Maximum number of concurrent reconcile loops")
-	fs.BoolVar(&cfg.EnableCloudWatchLogs, flagEnableCloudWatchLogs, false, "If enabled, policy decision logs will be streamed to CloudWatch")
+	fs.BoolVar(&cfg.EnablePolicyEventLogs, flagEnablePolicyEventLogs, false, "If enabled, policy decision logs will be collected & logged")
```
achevuru, line 45: Flow specific logs will be rather useful for debugging. Since this is an initial release, will be helpful to have it turned on by default for the first few releases. It should only log an entry per flow and not per packet (for accepted flows). Obviously, this change allows an user to turn it off if it becomes too verbose for them..

mycrEEpy: The problem with turning it on by default is not the amount of logs being produced, but the fact the the busy loop which is collecting the decision logs is consuming a full CPU core on every node. From a users point of view this unexpected and undesired default behavior.

achevuru: Sorry for the delay. We had some internal discussions on what the default should be for this flag and we agreed on disabling this by default considering the agent runs as a daemonset. So, we should be good with the current PR. We will look to enable it by default, once we find a way to optimize the CPU usage. Thanks for the PR.

achevuru (approved):

achevuru:
Can you help rebase the PR?

mycrEEpy:
> Can you help rebase the PR?

I've rebased onto main.

jayanthvn:
We need to block init of Ring buffer and epoll... I will take up post merge of this PR.

achevuru (approved):

PR #57, opened 2023-09-08
Update scripts to run cyclonus suite and install latest MAO

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
Adds a script to run cyclonus tests on an existing cluster and updates the MAO scripts to install the latest managed addon for vpc-cni instead of specifying the version

### DIFF
--- a/.github/workflows/performance-tests.yaml
+++ b/.github/workflows/performance-tests.yaml
@@ -31,10 +31,9 @@ jobs:
       - name: Run performance tests
         env:
           RUN_PERFORMANCE_TESTS: true
-          KUBERNETES_VERSION: 1.27
+          K8S_VERSION: 1.27
           NODES_CAPACITY: 3
           INSTANCE_TYPE: c5.xlarge
-          CNI_ADDON_VERSION: v1.14.0-eksbuild.3
           IP_FAMILY: ${{ matrix.ip-family }}
         run: |
           ./scripts/run-tests.sh
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -5,9 +5,10 @@ This package contains shell scripts and libraries used for running e2e integrati
 ### run-test.sh
 
 `run-test.sh` can run various integration test suites against the current revision in the invoking directory. 
+`run-cyclonus-tests.sh` Runs cyclonus tests against an existing cluster and validates the output
 
 #### Tests
-The following tests are valid to run, and setting the respective environment variable to true will run them:
+The following tests are valid to run using `run-test.sh` script, and setting the respective environment variable to true will run them:
 1. Conformance Tests - `RUN_CONFORMANCE_TESTS`
 2. Performance Tests - `RUN_PERFORMANCE_TESTS`
 
--- a/scripts/lib/cleanup.sh
+++ b/scripts/lib/cleanup.sh
@@ -6,19 +6,21 @@ function check_path_cleanup(){
     do
         export NODE=$node
         envsubst '$NODE' < ${DIR}/test/check-cleanup-pod.yaml > ${DIR}/test/check-cleanup-pod-$node.yaml
-        kubectl apply -f ${DIR}/test/check-cleanup-pod-$node.yaml
+        kubectl apply -f ${DIR}/test/check-cleanup-pod-$node.yaml -n default
         rm -rf ${DIR}/test/check-cleanup-pod-$node.yaml
     done
     sleep 20
 
     for node in $worker_nodes
     do
-        if [[ $(kubectl get pods $node -ojsonpath="{.status.phase}") == "Failed" ]]; then
-            echo "BPF files not cleaned up on $node.. $(kubectl logs $node)"
-            exit 1
+        if [[ $(kubectl get pods -n default $node -ojsonpath="{.status.phase}") == "Failed" ]]; then
+            echo "BPF files not cleaned up on $node"
+            kubectl logs $node -n default
+            TEST_FAILED=true
+        else
+            echo "BPF files were cleaned up from the node $node"
         fi
-        kubectl delete pods $node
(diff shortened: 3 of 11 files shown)


scripts/README.md
```diff
@@ -5,9 +5,10 @@ This package contains shell scripts and libraries used for running e2e integrati
 ### run-test.sh
 
 `run-test.sh` can run various integration test suites against the current revision in the invoking directory. 
+`run-cyclonus-tests.sh` Runs cyclonus tests against an existing cluster and validates the output
 
 #### Tests
-The following tests are valid to run, and setting the respective environment variable to true will run them:
+The following tests are valid to run using `run-test.sh` script, and setting the respective environment variable to true will run them:
```
jayanthvn, line 11: Maybe can add a make target for this?

scripts/lib/cleanup.sh
```diff
@@ -6,19 +6,21 @@ function check_path_cleanup(){
...
-            exit 1
+        if [[ $(kubectl get pods -n default $node -ojsonpath="{.status.phase}") == "Failed" ]]; then
+            echo "BPF files not cleaned up on $node"
+            kubectl logs $node -n default
+            TEST_FAILED=true
+        else
+            echo "BPF files were cleaned up from the node $node"
         fi
-        kubectl delete pods $node
+        kubectl delete pods $node -n default
```
jayanthvn, line 23: Btw, we tear down the nodes so it is fine, if not cyclonus last test will leak some resources such as pods and ns..

jaydeokar: We use the `--cleanup-namespaces=true` in cyclonus tests so ideally if the job finishes, it ends up cleaning the namespaces that it had created.

scripts/README.md
jaydeokar: Good call, I'll update this

.github/workflows/performance-tests.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Run performance tests
         env:
           RUN_PERFORMANCE_TESTS: true
-          KUBERNETES_VERSION: 1.27
+          K8S_VERSION: 1.27
           NODES_CAPACITY: 3
           INSTANCE_TYPE: c5.xlarge
           CNI_ADDON_VERSION: v1.14.0-eksbuild.3
```
jayanthvn, line 37: Maybe even for this we can pick the latest add-on?

scripts/README.md
```diff
@@ -5,9 +5,10 @@ This package contains shell scripts and libraries used for running e2e integrati
 ### run-test.sh
 
 `run-test.sh` can run various integration test suites against the current revision in the invoking directory. 
+`run-cyclonus-tests.sh` Runs cyclonus tests against an existing cluster and validates the output
```
jayanthvn, line 8: Not needed now but for the next PR, we can add a test or another workflow for disabling controller in CP and installing the controller in DP followed by conformance tests.

scripts/lib/cleanup.sh
```diff
@@ -6,19 +6,21 @@ function check_path_cleanup(){
...
         rm -rf ${DIR}/test/check-cleanup-pod-$node.yaml
     done
     sleep 20
 
     for node in $worker_nodes
     do
-        if [[ $(kubectl get pods $node -ojsonpath="{.status.phase}") == "Failed" ]]; then
-            echo "BPF files not cleaned up on $node.. $(kubectl logs $node)"
-            exit 1
+        if [[ $(kubectl get pods -n default $node -ojsonpath="{.status.phase}") == "Failed" ]]; then
```
jayanthvn, line 16: Trying to understand this, are we checking the pod status to determine if the BPF files are cleaned up? Am I missing anything? We should check on the node in /sys/fs/bpf and see if there are any leaked maps/progs or we could also just dump the NA CLI to check. But after cleanup we should have just 2 maps i.e, global map - ringbuffer and conntrack map. There will be 2 additional maps which is loaded by default which we can ignore.

scripts/lib/cluster.sh
```diff
@@ -1,20 +1,18 @@
...
+function load_default_values(){
 
     CLUSTER_NAME=network-policy-${RANDOM}
     : "${AWS_REGION:=us-west-2}"
     : "${AMI_FAMILY:=AmazonLinux2}"
     : "${NODEGROUP_TYPE:=linux}"
     : "${NODES_CAPACITY:=3}"
     : "${INSTANCE_TYPE:=t3.large}"
-    : "${KUBERNETES_VERSION:=1.27}"
+    : "${K8S_VERSION:=1.27}"
```
jayanthvn, line 11: Soon we might have to update this to 1.28...

scripts/lib/cleanup.sh
jaydeokar: Yes, the pod returns a failure if either the maps/programs are not cleaned up (except the two global maps). The pod mounts the host file system and runs the check if there are any leaked resources on every instance.

.github/workflows/performance-tests.yaml
jaydeokar: Updated

scripts/README.md
jaydeokar: Ack, I'll keep it in my backlog

scripts/README.md
```diff
@@ -5,9 +5,10 @@ This package contains shell scripts and libraries used for running e2e integrati
 ### run-test.sh
 
 `run-test.sh` can run various integration test suites against the current revision in the invoking directory.
```
jayanthvn, line 7: Since this is used only for GH actions, we can add a note here..make targets can be added as a follow up..

jayanthvn (approved):
lgtm

Issue #56, opened 2023-09-08
Liveness Probes (from host kubelet) are incorrectly blocked


vaskozl:
Upon installing [flux](https://github.com/fluxcd/flux2) with the default network policies, the flux controllers will start CrashLoopBackoffing as the liveness probes are blocked by the controller.

Host to local workload traffic should not be blocked by network policies and is allowed by other CNIs for the purpose of readiness/liveness probes. Furthermore there is no way to know what the IP of the host node is, so to make liveness probes work you'd have to probably allow the entire subnet, which defeats the goal of most ingress policies.

Note that at present some pod's the liveness probe may not fail as netpols can fail to apply completely with the latest release. This may be fixed by https://github.com/aws/aws-network-policy-agent/commit/7e61c86bc5589e2f4866fd45a2b7c16737c88635.

achevuru:
@vaskozl Traffic to/from Node IP will be allowed by default (i.e.,) agent will derive the node's Host IP from IMDS and will automatically add it to the list of allowed CIDRs. Let me validate that and get back..

Also, is this on a v4 or v6 cluster?

vaskozl:
My testing has been on a v6 cluster. The policies can be found on the flux repo:
https://github.com/fluxcd/flux2/tree/9b1e160798220dfe8d79bc72e2bd75d9f3de45eb/manifests/policies

Or generated with `flux install --export`.

achevuru:
@vaskozl  Validated that (http based) liveness/readiness probes work fine for pods with active Network policies in both IPv4 and IPv6 clusters.

Issue with flux is due to `allow-scraping` policy selecting all the pods on the cluster on ingress and allowing them only for port 8080 and TCP. This policy selects host networking pods running on the cluster as well (For ex: kube-proxy and aws-node pods) and host networking pods share the same IP as the node IP.

Although, we allow ingress and egress traffic to node IP by default across all ports and protocols, the above policy resulted in narrowing down the allowed port + protocol combo to 8080 and TCP for Node IP in ingress for these pods. Flux pods have http probes configured for liveness/readiness but on ports 9xxx and so the liveness/readiness probes were rejected as kubelet will use the node IP as the source IP for the probes. We will address the issue in the agent to account for conflicts between policies selecting host networking pods and node IP.

jayanthvn:
PR is merged, will close the issue.

PR #58, opened 2023-09-12
With catchALL honor "except"

jayanthvn:
*Issue #, if available:* N/A

*Description of changes:*

A network policy can allow catchALL i.e, 0.0.0.0/0 and block certain cidr which was getting ignored for catch all entries.

Sample policy

```
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  name: test-np
  namespace: default
spec:
  podSelector:
    matchLabels:
      app: nginx
  egress:
  - to:
    - ipBlock:
        cidr: 0.0.0.0/0
        except:
          - 172.31.54.64/32
  policyTypes:
    - Egress
```

Verdict is `Accept` ->

```
{"level":"info","timestamp":"2023-09-12T17:36:37.991Z","logger":"ebpf-client","msg":"Flow Info:  ","Src IP":"192.168.21.79","Src Port":0,"Dest IP":"172.31.54.64","Dest Port":0,"Proto":"ICMP","Verdict":"ACCEPT"}
```

After fix...

Except is honored ->

```
{"level":"info","timestamp":"2023-09-12T17:41:30.922Z","logger":"ebpf-client","msg":"Total L4 entry count for catch all entry: ","count: ":0}
{"level":"info","timestamp":"2023-09-12T17:41:30.922Z","logger":"ebpf-client","msg":"L4 values: ","protocol: ":254,"startPort: ":0,"endPort: ":0}
{"level":"info","timestamp":"2023-09-12T17:41:30.922Z","logger":"ebpf-client","msg":"Parsed Except CIDR","IP Key: ":"172.31.54.64/32"}
{"level":"info","timestamp":"2023-09-12T17:41:30.922Z","logger":"ebpf-client","msg":"L4 values: ","protocol: ":255,"startPort: ":0,"endPort: ":0}
```

Verdict is `Deny` ->
```
{"level":"info","timestamp":"2023-09-12T17:44:04.821Z","logger":"ebpf-client","msg":"Flow Info:  ","Src IP":"192.168.21.79","Src Port":0,"Dest IP":"172.31.54.64","Dest Port":0,"Proto":"ICMP","Verdict":"DENY"}
{"level":"info","timestamp":"2023-09-12T17:44:04.821Z","logger":"ebpf-client","msg":"Sending logs to CW"}
[...]

PR #65, opened 2023-09-14
Ignore policy restrictions against Node IP

achevuru:
*Issue #, if available:*
#56 

*Description of changes:*

Network Policy Agent by default allows unrestricted access to/from Node IP. However, if a Network policy selects host networking pods (that are assigned node IP) and narrows down the allowed list of port and protocol combinations, the agent currently honors that but that might break the (http) liveness/readiness probes unless the policy explicitly allows the probe endpoint ports. PR will address this issue and will make sure the enforced policies allow all traffic to/from Node IP.

https://kubernetes.io/docs/concepts/services-networking/network-policies/#what-you-can-t-do-with-network-policies-at-least-not-yet

```
Pods cannot currently block localhost access, nor do they have the ability to block access from their resident node
```

### DIFF
--- a/pkg/ebpf/bpf_client.go
+++ b/pkg/ebpf/bpf_client.go
@@ -734,8 +734,9 @@ func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirew
 		if !strings.Contains(string(firewallRule.IPCidr), "/") {
 			firewallRule.IPCidr += v1alpha1.NetworkAddress(l.hostMask)
 		}
-		//TODO - Just Purge both the entries and avoid these calls for every CIDR
-		if utils.IsCatchAllIPEntry(string(firewallRule.IPCidr)) {
+
+		if utils.IsCatchAllIPEntry(string(firewallRule.IPCidr)) ||
+			utils.IsNodeIP(l.nodeIP, string(firewallRule.IPCidr)) {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/ebpf/bpf_client.go
```diff
@@ -734,8 +734,9 @@ func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirew
 		if !strings.Contains(string(firewallRule.IPCidr), "/") {
 			firewallRule.IPCidr += v1alpha1.NetworkAddress(l.hostMask)
 		}
-		//TODO - Just Purge both the entries and avoid these calls for every CIDR
-		if utils.IsCatchAllIPEntry(string(firewallRule.IPCidr)) {
+
+		if utils.IsCatchAllIPEntry(string(firewallRule.IPCidr)) ||
```
jayanthvn, line 738: I have removed this check as part of https://github.com/aws/aws-network-policy-agent/pull/58 but once we merge this PR, I will just keep the isNodeIP check..

pkg/utils/utils.go
```diff
@@ -219,6 +219,14 @@ func IsCatchAllIPEntry(ipAddr string) bool {
 	return false
 }
 
+func IsNodeIP(nodeIP string, ipCidr string) bool {
```
jayanthvn, line 222: Is there any special handle needed for IPv6 compact format?

achevuru: net.Parse* functions should take care of that

jayanthvn (approved):
lgtm

PR #106, opened 2023-10-24
Handle PolicyEndpoint split scenario when the target pods are paired …

achevuru:
…with empty ingress/egress rules
*Description of changes:*

Network Policy controller that runs in EKS Control Plane is responsible for resolving the Pod and Namespace selectors specified in a Network Policy resource and updates the resolved endpoints in a custom resource called PolicyEndpoints . When the number of endpoints across ingress, egress or pod selectors breach a certain threshold, it splits them in to multiple PolicyEndpoints resources.  Approach is similar to how endpoint slices are implemented in Kubernetes. It is possible to land in a scenario where targeted pods are paired with either only a subset of ingress/egress rules or none at all in the corresponding `PolicyEndpoint` resource. PR aims to address this scenario.


controllers/policyendpoints_controller.go
```diff
@@ -418,11 +418,44 @@ func (r *PolicyEndpointsReconciler) updateeBPFMaps(ctx context.Context, podIdent
 	return nil
 }
 
+func (r *PolicyEndpointsReconciler) deriveTargetPodsForParentNP(ctx context.Context,
+	policyEndpoint *policyk8sawsv1.PolicyEndpoint) ([]types.NamespacedName, map[string]bool, []types.NamespacedName) {
+	var targetPods, podsToBeCleanedUp []types.NamespacedName
+	podIdentifiers := make(map[string]bool)
+	currentPE := &policyk8sawsv1.PolicyEndpoint{}
+
+	r.log.Info("Parent NP resource:", "Name: ", policyEndpoint.Spec.PolicyRef.Name)
```
jayanthvn, line 427: Do we need to maybe add a `nil` check? Even in the caller reconcildePolicyEndpoint we don't have the check.

controllers/policyendpoints_controller.go
```diff
@@ -547,3 +592,23 @@ func (r *PolicyEndpointsReconciler) getLocalConntrackCacheCleanupPeriod() time.D
 	}
 	return defaultLocalConntrackCacheCleanupPeriodInSeconds
 }
+
+func (r *PolicyEndpointsReconciler) derivePolicyEndpointsOfParentNP(ctx context.Context, parentNP, resourceNamespace string) []string {
+	var parentPolicyEndpointList []string
```
jayanthvn, line 597: Nit: Maybe just `policyEndpointList`?

controllers/policyendpoints_controller.go
jayanthvn: Synced up offline and reconcile ensures the object is initialized.

jayanthvn (approved):
lgtm

Issue #135, opened 2023-11-09
Network Policy Option Cannot Disable


atilsensalduz:
Hey team,

I encountered an error when attempting to set the 'enableNetworkPolicy' parameter to false. Could you please guide me on the correct procedure to disable network policies?

Error logs on pod:

```
{"level":"info","ts":"2023-11-09T12:32:26.065Z","caller":"runtime/asm_arm64.s:1197","msg":"version","GitVersion":"","GitCommit":"","BuildDate":""}
2023-11-09 12:32:26.067401356 +0000 UTC Logger.check error: failed to get caller
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xaaaae21acddc]

goroutine 84 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/internal/controller/controller.go:116 +0x1a4
panic({0xaaaae2e658e0?, 0xaaaae4298440?})
	/root/sdk/go1.21.3/src/runtime/panic.go:914 +0x218
github.com/aws/aws-network-policy-agent/controllers.(*PolicyEndpointsReconciler).configureeBPFProbes(0x4000150540, {0xaaaae31efe58, 0x40009250b0}, {0x4000c07ec0, 0x25}, {0x4000936aa0?, 0x1, 0x0?}, {0x400092bd80, 0x1, ...}, ...)
	/workspace/controllers/policyendpoints_controller.go:258 +0x34c
github.com/aws/aws-network-policy-agent/controllers.(*PolicyEndpointsReconciler).reconcilePolicyEndpoint(0x4000150540, {0xaaaae31efe58, 0x40009250b0}, 0x400093e9c0)
	/workspace/controllers/policyendpoints_controller.go:232 +0x550
github.com/aws/aws-network-policy-agent/controllers.(*PolicyEndpointsReconciler).reconcile(0x4000150540, {0xaaaae31efe58, 0x40009250b0}, {{{0x40004f5600, 0xc}, {0x400015c858, 0x15}}})
	/workspace/controllers/policyendpoints_controller.go:149 +0x1a4
[...]

PR #143, opened 2023-11-23
disable reconciliation when enable-network-policy flag is set

larhauga:
Issue #135

Description of changes: When `enable-network-policy` flag is false, the ebpfClient is not configured, which results in a nilpointer. This change breaks the reconcile cycle to ensure that the client is not used. This will however result in a state that is not desirable, but more recoverable than having a crashing cni driver.

The intent was to make a minimal intrusive change to mitigate the nilpointers that happen if the driver has been started with network policies, before it has been disabled again (rollback scenarios).


achevuru:
@larhauga  Thanks for the PR. Our intent behind the current approach is to fail hard (crash) when there are active `PolicyEndpoint` resources in the cluster with `NetworkPolicy` feature disabled. Behavior can be undefined in such scenarios as there can be lingering stale probes against currently active pods on the cluster. So, essentially this is a misconfiguration that can lead to traffic issues and we want the user to act on it before reverting to the disabled mode.

It also points to a scenario where [Network Policy controller](https://github.com/aws/amazon-network-policy-controller-k8s/) running in EKS control plane is actively reconciling on Network policy resources in the cluster and updating the relevant `policyEndpoint` resources and this is problematic in few ways...

- Network Policy controller will unnecessarily consume CPU and Mem resources of EKS control plane instances trying to resolve the NPs, even though the feature is disabled in VPC CNI.
- Might be problematic if an user wants to use a 3rd party Network Policy solution instead and we have stale probes lingering around on the nodes.

Issue #150, opened 2023-11-29
aws-eks-na-cli loaded-ebpfdata fails with unable to get FD


fungusakafungus:
**What happened**:
`/opt/cni/bin/aws-eks-na-cli ebpf loaded-ebpfdata` fails with
> Failed to execute the cmd -  failed walking the bpfdirectory unable to get FD

**Attach logs**
```
[root@ip-10-5-90-7 ~]# date
Wed Nov 29 17:24:02 UTC 2023

[root@ip-10-5-90-7 ~]# /opt/cni/bin/aws-eks-na-cli ebpf loaded-ebpfdata
Failed to execute the cmd -  failed walking the bpfdirectory unable to get FD

[root@ip-10-5-90-7 ~]# cat /var/log/aws-routed-eni/ebpf-sdk.log | grep 17:24: | jq .msg -r
Constructed new logger instance
Initialized new logger as an existing instance was not found
Initialized new SDK cache as an existing instance was not found
Dumping pinpaths - %!(EXTRA string=/sys/fs/bpf/globals/aws/maps/busybox-default_egress_map)
Got ID 8
Found Identified - busybox-default : egress_map
Found ->  %!(EXTRA string=busybox-default, string=egress_map)
Adding egress_map -> %!(EXTRA string=busybox-default)
Found Identified - busybox-default : egress_map
Found ->  %!(EXTRA string=busybox-default, string=egress_map)
Adding egress_map -> %!(EXTRA string=busybox-default)
Got FD 5
Adding ID 8 to name egress_map and NS busybox-default
Dumping pinpaths - %!(EXTRA string=/sys/fs/bpf/globals/aws/maps/busybox-default_ingress_map)
Got ID 7
Found Identified - busybox-default : ingress_map
Found ->  %!(EXTRA string=busybox-default, string=ingress_map)
Adding ingress_map -> %!(EXTRA string=busybox-default)
Found Identified - busybox-default : ingress_map
Found ->  %!(EXTRA string=busybox-default, string=ingress_map)
Adding ingress_map -> %!(EXTRA string=busybox-default)
Got FD 8
Adding ID 7 to name ingress_map and NS busybox-default
Dumping pinpaths - %!(EXTRA string=/sys/fs/bpf/globals/aws/maps/global_aws_conntrack_map)
Got ID 5
Found Identified - global : aws_conntrack_map
Found ->  %!(EXTRA string=global, string=aws_conntrack_map)
[...]

PR #152, opened 2023-11-30
CLI changes

jayanthvn:
*Issue #, if available:* #150 

*Description of changes:*
Walk for cli was skipping because we used to check on SDK cache and that will be available only during recovery.

Additionally port/protocol is an array and as part of this we will print all ports/protocols

```
*******************************
Key : IP/Prefixlen - 192.168.81.90/32 
-------------------
Value Entry :  0
Protocol -  TCP
StartPort -  8080
Endport -  0
-------------------
-------------------
Value Entry :  1
Protocol -  ANY PROTOCOL
StartPort -  8080
Endport -  0
-------------------
```

```
PinPath:  /sys/fs/bpf/globals/aws/programs/source-controller-56ccbf8db8-flux-system_handle_ingress
Pod Identifier : source-controller-56ccbf8db8-flux-system  Direction : ingress 
Prog ID:  151
Associated Maps -> 
Map Name:  aws_conntrack_map
Map ID:  5
Map Name:  ingress_map
Map ID:  149
Map Name:  policy_events
Map ID:  6
```

This needs SDK release so making it a draft for now...


jayanthvn:
Pending SDK release..

achevuru (approved):

PR #159, opened 2023-12-05
Ignore PE slices tied to same NP during Clean up flow

achevuru:
*Description of changes:*

Agent will leave the probes in place for pods if there are other active PE resources against them but we can/should ignore if the PEs are tied to the same parent NP and proceed to detach the probes in these scenarios.


jayanthvn (approved):
lgtm

jayanthvn (approved):

Issue #103, opened 2023-10-19
UTC Logger.check error: failed to get caller


tl-alex-nicot:
looks like with v1.15.1 of the aws vpc cni the output for the aws-eks-nodeagent just says

```
k logs aws-node-jpg2r -c aws-eks-nodeagent
{"level":"info","ts":"2023-10-19T09:55:41.395Z","caller":"runtime/asm_amd64.s:1598","msg":"version","GitVersion":"","GitCommit":"","BuildDate":""}
2023-10-19 09:55:41.481551956 +0000 UTC Logger.check error: failed to get caller
2023-10-19 09:55:41.481656868 +0000 UTC Logger.check error: failed to get caller
```

jayanthvn:
It shouldn't cause any functionality impact but we will clean it up.

Jufik:
Hey! 

Using VPC CNI v1.16.0 I'm still facing this Issue.
This makes enable-policy-event-logs pretty useless and the logging feature pretty useless.
Cluster uses EKS (1.27) with managed node groups (1.27.9-20240117): any chance to get a pointer on how to fix that and get  hands on policy event logs ?

jayanthvn:
@Jufik - Are you checking pod logs for policy logs when you enable - `enable-policy-event-logs` ? The decision/access logs will be redirected to `/var/log/aws-routed-eni/network-policy-agent.log` when the knob is enabled. The same log file will also contain node agent logs..

snarfmonkey:
Am i doing something wrong, because I am using 1.16.3 and still see `Logger.check error: failed to get caller` logged constantly when policy event logs are enabled. I'd rather not log redundant, unimportant lines on every since pod on each node. Is is possible to change the log level so I don't have to log these to disk at all?

atilsensalduz:
The problem persists, and I'm encountering it with aws-network-policy-agent:v1.0.8-eksbuild.1 and eks 1.27. Can we consider reopening the issue? @jayanthvn @jdn5126 

Additionally, when I enable both the enable-cloudwatch-logs and enable-policy-event-logs parameters, the pods get stuck in a crashloopbackoff state with exit code 1, and no logs are generated.

giedriuskilcauskas:
I can confirm that this issue still persists with EKS 1.28 / aws-network-policy-agent:v1.1.0-eksbuild.1
Can this issue be /reopen 'ed?

parvez99:
[...]

Issue #49, opened 2023-09-05
Build info in logs is empty


sidewinder12s:
On startup it looks like this info is all empty: 

```log
{"level":"info","ts":"2023-09-05T21:59:11Z","msg":"version","GitVersion":"","GitCommit":"","BuildDate":""}
```

jdn5126:
Closing as fixed by #168

PR #168, opened 2023-12-26
Fix metrics logging error; Remove version log

jdn5126:
*Issue #, if available:*
https://github.com/aws/aws-network-policy-agent/issues/103
https://github.com/aws/aws-network-policy-agent/issues/49

*Description of changes:*
This PR fixes two issues:
1. The metrics server runs in a goroutine. Logging was broken in the metrics server because the call stack was starting at -2 when it needed to start at 0. This PR instantiates a new logger in the metrics goroutine and resolves https://github.com/aws/aws-network-policy-agent/issues/103
2. During init, the agent was trying to log version information. This information is not properly set, leading to https://github.com/aws/aws-network-policy-agent/issues/49. This PR removes that log, as the image tag is sufficient to determine version.

New log looks like:
```
% kubectl logs -n kube-system aws-node-57t4f -c aws-eks-nodeagent
{"level":"info","ts":"2023-12-26T20:57:29.306Z","caller":"metrics/metrics.go:23","msg":"Serving metrics on ","port":61680}
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/aws/aws-network-policy-agent/pkg/logger"
 
-	"github.com/aws/aws-network-policy-agent/pkg/version"
 	"github.com/go-logr/logr"
 	"github.com/go-logr/zapr"
 	"github.com/spf13/pflag"
@@ -56,11 +55,6 @@ func init() {
 
 func main() {
 	initLogger, _ := getLoggerWithLogLevel("info", "")
-	initLogger.Info("version",
-		"GitVersion", version.GitVersion,
-		"GitCommit", version.GitCommit,
-		"BuildDate", version.BuildDate,
-	)
 
 	ctrlConfig, err := loadControllerConfig()
 	if err != nil {
@@ -111,7 +105,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	go metrics.ServeMetrics(setupLog)
+	go metrics.ServeMetrics()
 
 	setupLog.Info("starting manager")
 	if err := mgr.Start(ctx); err != nil {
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -56,11 +55,6 @@ func init() {
 
 func main() {
 	initLogger, _ := getLoggerWithLogLevel("info", "")
-	initLogger.Info("version",
-		"GitVersion", version.GitVersion,
```
jayanthvn, line 60: Yeah we never rebuild the same tag so just the tag should be sufficient to identify the commit..

jayanthvn (approved):
lgtm

main.go
```diff
@@ -135,6 +129,6 @@ func loadControllerConfig() (config.ControllerConfig, error) {
 
 // getLoggerWithLogLevel returns logger with specific log level.
 func getLoggerWithLogLevel(logLevel string, logFilePath string) (logr.Logger, error) {
-	ctrlLogger := logger.New(logLevel, logFilePath)
+	ctrlLogger := logger.New(logLevel, logFilePath, 2)
```
jayanthvn, line 132: Nit: later we can move the logger defaults of 0,1,2..to utils as enum

pkg/metrics/metrics.go
```diff
@@ -43,3 +45,8 @@ func setupMetricsServer() *http.Server {
 	}
 	return server
 }
+
+func getMetricsLogger() logr.Logger {
+	ctrlLogger := logger.New("info", "", 0)
```
jayanthvn, line 50: Nit: Same as above..this can be a util enum..but can take it up as followup..

main.go
jdn5126: Yeah, I considered an enum, but could not come up with any names that make sense, as the value to use is dependent on where the logger will be called from. If you are within a goroutine function, you want 0, as above is the standard library runtime. If you are within the controller handle routines, you want 2, as you want to skip the controller call stack surrounding the function

PR #184, opened 2024-01-17
Check the new addon versions in the right regions

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
We should look for addon versions in the right regions instead of looking in the default region where the test is run from


orsenthil (approved):
LGTM

Issue #109, opened 2023-10-24
please remove python2.7 from the network-policy agent


maxres-ch:
**What would you like to be added**:
I don't want anything added, I want something removed.
Current python2.7 is the network policy agent.
```
❯ k exec -it -n kube-system aws-node-wt6mn -c aws-eks-nodeagent -- bash
bash-4.2# python2.7
Python 2.7.18 (default, Feb 28 2023, 02:51:06)
[GCC 7.3.1 20180712 (Red Hat 7.3.1-15)] on linux2
Type "help", "cop
```
**Why is this needed**:

We run in a regulated environment python2.7 has open CVEs that wont be resolved and now we have audit and infosec demanding that we fix the CVEs.

jdn5126:
Transferred to Network Policy agent repo

jdn5126:
@maxres-ch this comes from the AL2 base image, I assume: https://github.com/aws/aws-network-policy-agent/blob/main/Dockerfile#L32

I don't see an issue with uninstalling python2.7, as it is not actually used

jayanthvn:
This is a duplicate of #107.

maxres-ch:
@jdn5126 should I open a PR?

jdn5126:
> @jdn5126 should I open a PR?

sure, that would be great, thank you! You are just planning to yum uninstall it?

maxres-ch:
> You are just planning to yum uninstall it? 

I was. I don't `yum` much but, I'll try to see how deps can be uninstalled, how that does feel a bit dangerous I'm not sure what kind of existing test coverage that would catch it there.

FWIW I might do this via my personal account @grumps.

grumps:
@jdn5126 I looked into this. It looks like the image is based on AL2 and within it yum requires python2.7.  It can't be removed until the base image moves to AL 2023.

maxres-ch:
Any chance you know when/if https://github.com/aws/aws-network-policy-agent/blob/main/Dockerfile#L44

Will move to AL 2023, there's no python2.7 here.

jdn5126:
I see, thank you for the investigation. I did not realize that AL2 had a hard dependency on Python 2.7. I am not sure about the feasibility of the Network Policy agent moving to AL2023 or to a minimum image build. @jayanthvn @achevuru any thoughts here?

jdn5126:
[...]

PR #212, opened 2024-02-15
Use minimal base image for Network Policy agent; update Golang version to 1.21.7

jdn5126:
*Issue #, if available:*
https://github.com/aws/aws-network-policy-agent/issues/109

*Description of changes:*
This PR modifies the Network Policy agent Dockerfile to use `eks-distro-minimal-base-glibc` as the base image. This minimal image, along with removing the libraries that were installed using `yum`, significantly decreases the size of the container image. My personal build showed a decrease from ~290MB to ~38MB.

A smaller container image means faster image pulls and a smaller surface area for CVEs.

For the libraries that were installed in the container filesystem using `yum`, such as `libbpf-devel`, it turns out that none of these are needed at runtime. The object files created in the `bpfbuilder` stage already have everything they need statically linked.

Also, I updated the Golang builder image to Golang 1.21.7.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM public.ecr.aws/eks-distro-build-tooling/golang:1.21.5-6-gcc-al2 as builder
+FROM public.ecr.aws/eks-distro-build-tooling/golang:1.21.7-8-gcc-al2 as builder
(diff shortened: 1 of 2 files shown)


jayanthvn (approved):
lgtm..

Issue #118, opened 2023-10-30
Network Policy agent reconciler crashes with a runtime error when pod name contains dot(.) .


ibnjunaid:
**What happened:**
Network Policy agent crashes when pod name contains . (dot) with the following runtime error.
```
Observed a panic in reconciler: runtime error: invalid memory address or nil pointer dereference
```

Attached logs:
```log
NAMESPACE     NAME                              READY   STATUS             RESTARTS      AGE
default       sample-app.app-7c547489fb-gmnj9   1/1     Running            0             3m46s
kube-system   aws-node-nmxkh                    1/2     CrashLoopBackOff   5 (54s ago)   4m23s
kube-system   coredns-664b6f5f5c-9kq9w          1/1     Running            0             46h
kube-system   coredns-664b6f5f5c-jhzmm          1/1     Running            0             46h
kube-system   kube-proxy-7g9qg                  1/1     Running            0             5h1m
```

```
 containerStatuses:
  - containerID: containerd://62da97335e19f10f460f78b12dea2a60e0a8e3938a6a7e5f38b7e72e00fff4cf
    image: 602401143452.dkr.ecr.us-east-2.amazonaws.com/amazon/aws-network-policy-agent:v1.0.2-eksbuild.1
    imageID: 602401143452.dkr.ecr.us-east-2.amazonaws.com/amazon/aws-network-policy-agent@sha256:71fbb862ba51217f4c8a22502cba6fa8baa098b80590ea753378694b7c82a4db
    lastState:
      terminated:
        containerID: containerd://62da97335e19f10f460f78b12dea2a60e0a8e3938a6a7e5f38b7e72e00fff4cf
        exitCode: 2
        finishedAt: "2023-10-30T09:55:23Z"
        reason: Error
        startedAt: "2023-10-30T09:55:23Z"
    name: aws-eks-nodeagent
    ready: false
    restartCount: 6
    started: false
    state:
      waiting:
        message: back-off 5m0s restarting failed container=aws-eks-nodeagent pod=aws-node-nmxkh_kube-system(ba6898d2-9c89-45e3-8202-2d97012e760f)
        reason: CrashLoopBackOff
```

```json
{"level":"info","timestamp":"2023-10-30T09:46:22.273Z","msg":"Starting EventSource","controller":"policyendpoint","controllerGroup":"networking.k8s.aws","controllerKind":"PolicyEndpoint","source":"kind source: *v1alpha1.PolicyEndpoint"}
[...]

PR #225, opened 2024-03-04
Support for network policies applied to pods with '.' in their pod name

zachdorame:
*Issue #, if available:*
https://github.com/aws/aws-network-policy-agent/issues/118

*Description of changes:*
Pods with `.` in their pod name will now have `.` changed to `dot`. 

*Testing:*
I applied `Sample Deployment` and `Sample Policy` from this comment: https://github.com/aws/aws-network-policy-agent/issues/118#issue-1967965139 to a cluster. With this change, I observed that the network policy reconciler was no longer crashing and that `Sample Policy` was being successfully applied:

network-policy-agent.log:
```
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controller/controller.go:316","msg":"Received a new reconcile request","req":{"name":"deny-all-ingress-gtr69","namespace":"default"}}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:127","msg":"Processing Policy Endpoint  ","Name: ":"deny-all-ingress-gtr69","Namespace ":"default"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:146","msg":"Parent NP resource:","Name: ":"deny-all-ingress"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:211","msg":"Found another PE resource for the parent NP","name":"deny-all-ingress-gtr69"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:146","msg":"Total PEs for Parent NP:","Count: ":1}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:146","msg":"Derive PE Object ","Name ":"deny-all-ingress-gtr69"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:146","msg":"Processing PE ","Name ":"deny-all-ingress-gtr69"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:211","msg":"Found a matching Pod: ","name: ":"sample-app.app-7c547489fb-h72pk","namespace: ":"default"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:211","msg":"Derived ","Pod identifier: ":"sample-appdotapp-7c547489fb-default"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:466","msg":"Current PE Count for Parent NP:","Count: ":1}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:146","msg":"Total number of PolicyEndpoint resources for","podIdentifier ":"sample-appdotapp-7c547489fb-default"," are ":1}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:146","msg":"Deriving Firewall rules for PolicyEndpoint:","Name: ":"deny-all-ingress-gtr69"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:146","msg":"Total no.of - ","ingressRules":0,"egressRules":0}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:223","msg":"Default Deny enabled on Ingress"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:127","msg":"No Egress rules and no egress isolation - Appending catch allentry"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:146","msg":"Processing Pod: ","name:":"sample-app.app-7c547489fb-h72pk","namespace:":"default","podIdentifier: ":"sample-appdotapp-7c547489fb-default"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"ebpf-client","caller":"controllers/policyendpoints_controller.go:243","msg":"AttacheBPFProbes for","pod":"sample-app.app-7c547489fb-h72pk"," in namespace":"default"," with hostVethName":"enif41a4bc0fa2"}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"ebpf-client","caller":"controllers/policyendpoints_controller.go:270","msg":"Found an existing instance, let's derive the ingress context.."}
{"level":"info","ts":"2024-03-04T21:15:37.487Z","logger":"ebpf-client","caller":"controllers/policyendpoints_controller.go:270","msg":"Attempting to do an Ingress Attach ","with progFD: ":14}
{"level":"info","ts":"2024-03-04T21:15:37.509Z","logger":"ebpf-client","caller":"controllers/policyendpoints_controller.go:243","msg":"Successfully attached Ingress TC probe for","pod: ":"sample-app.app-7c547489fb-h72pk"," in namespace":"default"}
{"level":"info","ts":"2024-03-04T21:15:37.509Z","logger":"ebpf-client","caller":"controllers/policyendpoints_controller.go:270","msg":"Found an existing instance, let's derive the egress context.."}
{"level":"info","ts":"2024-03-04T21:15:37.509Z","logger":"ebpf-client","caller":"controllers/policyendpoints_controller.go:270","msg":"Attempting to do an Egress Attach ","with progFD: ":13}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"ebpf-client","caller":"controllers/policyendpoints_controller.go:243","msg":"Successfully attached Egress TC probe for","pod: ":"sample-app.app-7c547489fb-h72pk"," in namespace":"default"}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"controllers.policyEndpoints","caller":"controllers/policyendpoints_controller.go:146","msg":"Successfully attached required eBPF probes for","pod:":"sample-app.app-7c547489fb-h72pk","in namespace":"default"}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"ebpf-client","caller":"controllers/policyendpoints_controller.go:278","msg":"Pod has an Ingress hook attached. Update the corresponding map","progFD: ":14,"mapName: ":"ingress_map"}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"ebpf-client","caller":"ebpf/bpf_client.go:707","msg":"L4 values: ","protocol: ":254,"startPort: ":0,"endPort: ":0}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"ebpf-client","caller":"ebpf/bpf_client.go:707","msg":"Total L4 entry count for catch all entry: ","count: ":0}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"ebpf-client","caller":"controllers/policyendpoints_controller.go:436","msg":"ID of map to update: ","ID: ":98}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"ebpf-client","caller":"controllers/policyendpoints_controller.go:278","msg":"Pod has an Egress hook attached. Update the corresponding map","progFD: ":13,"mapName: ":"egress_map"}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"ebpf-client","caller":"ebpf/bpf_client.go:707","msg":"L4 values: ","protocol: ":254,"startPort: ":0,"endPort: ":0}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"ebpf-client","caller":"ebpf/bpf_client.go:707","msg":"Current L4 entry count for catch all entry: ","count: ":0}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"ebpf-client","caller":"ebpf/bpf_client.go:707","msg":"Total L4 entry count for catch all entry: ","count: ":0}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"ebpf-client","caller":"ebpf/bpf_client.go:707","msg":"L4 values: ","protocol: ":254,"startPort: ":0,"endPort: ":0}
{"level":"info","ts":"2024-03-04T21:15:37.525Z","logger":"ebpf-client","caller":"controllers/policyendpoints_controller.go:436","msg":"ID of map to update: ","ID: ":99}
```

Cyclonus tests also passed 


pkg/utils/utils.go
```diff
@@ -81,6 +82,20 @@ func GetPodIdentifier(podName, podNamespace string) string {
 	return podIdentifierPrefix + "-" + podNamespace
 }
 
+func addDots(podName string) string {
```
jdn5126, line 85: Can't we just use `strings.Replace(podName, ".", "_", -1)`?

pkg/utils/utils.go
```diff
@@ -81,6 +82,20 @@ func GetPodIdentifier(podName, podNamespace string) string {
...
+	var dotString strings.Builder
+
+	for _, c := range podName {
+		if c == '.' {
+			dotString.WriteString("dot")
```
jdn5126, line 90: I think replacing with the string "dot" makes this harder to read. Can we just replace with "_" instead? I guess we have to be careful about conflicting with existing pod names, but I do not see a way around that

jdn5126 (commented):
Regardless of how we replace the "." character, we want to make sure the controller has no issues when the name of the pod is changed. Did you try creating and deleting on the pod on a real cluster? And applying and removing network policies to the pod?

pkg/utils/utils.go
zachdorame: That's a lot cleaner, yeah we can

pkg/utils/utils.go
zachdorame: I see, yeah in upcoming revision it is "_"

zachdorame:
> Regardless of how we replace the "." character, we want to make sure the controller has no issues when the name of the pod is changed. Did you try creating and deleting on the pod on a real cluster? And applying and removing network policies to the pod?

I verified manually on a real cluster that I can create a pod `my.pod.with.dots` and that a network policy can be successfully applied and revoked. 
But, as part of my next revision, I've also added an integration test that performs this validation as well

test/integration/policy/policy_test.go
```diff
@@ -0,0 +1,94 @@
...
+	})
+
+})
+
+func processLogsForEndState(podlogs string, state string) error {
+
+	logSlice := strings.Split(strings.TrimSuffix(podlogs, "\n"), "\n")
+	endState := logSlice[len(logSlice)-1]
+
+	if endState != state {
```
jaydeokar, line 88: This will not work always as the last string will not always be "Success" / "Fail".

pkg/utils/utils.go
```diff
@@ -73,6 +73,7 @@ func GetPodNamespacedName(podName, podNamespace string) string {
 }
 
 func GetPodIdentifier(podName, podNamespace string) string {
+	podName = strings.Replace(podName, ".", "_", -1)
```
jdn5126, line 76: nit: since this can be confusing for a customer, I wonder if we should do something like:
```
if strings.Contains(podName, ".") {
   log.Info("Replacing '.' character with '_' for pod pin path.")
    podName = strings.Replace(podName, ".", "_", -1)
}
```

Also, I see that `GetPodIdentifier` is called in multiple places, like the RPC handler here: https://github.com/aws/aws-network-policy-agent/blob/main/pkg/rpc/rpc_handler.go#L50. We should confirm with @jayanthvn and @achevuru whether each caller needs the "." name or "_" name.

pkg/utils/utils_test.go
```diff
@@ -390,6 +390,34 @@ func TestGetPodIdentifier(t *testing.T) {
...
+		{
+			name: "Valid pod name and namespace",
+			args: args{
+				podName:      "my.pod.name-udp-748dc8d996-fb8b2",
+				podNamespace: "default",
+			},
+			want: "my_pod_name-udp-748dc8d996-default",
+		},
+	}
+	for _, tt := range tests {
```
jdn5126, line 413: nit: for completeness, let's do a negative test. So add a pod name without and "." and validate that it is not changed

jdn5126 (commented):
Integration test looks great! Just one nit about where we need to use the dot name and where we need to use the underscore name

test/integration/policy/policy_test.go
```diff
@@ -0,0 +1,94 @@
...
+		var clientNetworkPolicy *network.NetworkPolicy
+		var podName = "this.pod.has.dots.in.its.name"
+
+		BeforeEach(func() {
+			By("Creating a network policy that denies all egress traffic", func() {
+				clientNetworkPolicy = manifest.NewNetworkPolicyBuilder().
+					Namespace(namespace).
+					Name("deny-all-egress").
+					PodSelector("app", podName).
+					SetPolicyType(false, true).
```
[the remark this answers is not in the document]
zachdorame: I had tried that way but I found that the network policy kept getting set to an `Ingress` type. It turns out that I needed a way to explicitly set the policy to `Egress` since

> ...if you want to write a policy that specifies that no egress is allowed, you must specify a policyTypes value that include "Egress" (since such a policy would not include an egress section and would otherwise default to just [ "Ingress" ]).

from [NetworkPolicySpec](https://pkg.go.dev/k8s.io/api/networking/v1#NetworkPolicySpec)

test/integration/policy/policy_test.go
zachdorame: For my first test case where I apply a `deny-all-egress` policy and spin up the pod, I saw logs like...
```
Success
Success
Success
Fail
Fail
...
Fail
```
so the logs would start out as `Success`, and then the `deny-all-egress` network policy is applied to the pod, and the rest of the logs would become a long tail of `Fail`. 

I agree that if there is some perturbation of the system then the `Fail` could switch to `Success` and the test would seem to fail. 
One way to ensure that the test is genuinely detecting success or failure could be to use [MustPassRepeatedly()](https://onsi.github.io/ginkgo/#repeating-spec-runs-and-managing-flaky-specs), which will run the spec up to N times until they fail.

pkg/utils/utils_test.go
zachdorame: Will do

pkg/utils/utils.go
zachdorame: Ack, @achevuru had mentioned that `GetPodIdentifier()` would be a good place to perform the replacement but I'll still ping them for their final approval. and yeah I'll add a logging statement in there too

test/integration/policy/policy_test.go
zachdorame: Actually, I have an idea for how to make this work. For this test case I'll validate it with a state machine that expects the logs to begin with "Success" and to end with "Fail". I'll implement similar logic for the second policy removal spec too

pkg/utils/utils_test.go
zachdorame: Actually, this already existing test for `GetPodIdentifier` should be sufficient as a negative test: https://github.com/aws/aws-network-policy-agent/blob/main/pkg/utils/utils_test.go#L365

mjnovice:
Hye @zachdorame anything blocking this PR ? Would be happy to help if needed 🙂

achevuru (approved):

PR #254, opened 2024-04-17
Remove callSkip to prevent Logger.check error

ryota-sakamoto:
*Issue #, if available:*

#103

*Description of changes:*

The message `Logger.check error: failed to get caller` is generated by zap when call stack is empty after filtering callerSkip
https://github.com/uber-go/zap/blob/v1.27.0/logger.go#L382-L388

In previous PR https://github.com/aws/aws-network-policy-agent/pull/168, we addressed `failed to get caller` in the metrics server. But the error still exists because rpc handler has same issue.
For our use case, we don't need to use `zap.AddCallerSkip`. If we remove `zap.AddCallerSkip`, the message is disappeared and the log includes `caller`.

```
{"level":"info","ts":"2024-04-17T15:13:30.592Z","logger":"rpc-handler","caller":"rpc/rpc_handler.go:94","msg":"Serving RPC Handler","Address":"127.0.0.1:50052"}
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -133,6 +133,6 @@ func loadControllerConfig() (config.ControllerConfig, error) {
 
 // getLoggerWithLogLevel returns logger with specific log level.
 func getLoggerWithLogLevel(logLevel string, logFilePath string) (logr.Logger, error) {
-	ctrlLogger := logger.New(logLevel, logFilePath, 2)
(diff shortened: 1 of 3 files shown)


jayanthvn (approved):
lgtm

PR #259, opened 2024-04-24
Handle PolicyEndpoint Slice(s) deletion gracefully

achevuru:
*Description of changes:*

PR addresses scale down(/up) scenarios of `policyEndpoint` custom resource. Network Policy controller creates`policyEndpoint` resources with endpoint info that matches the selectors in the corresponding Network Policy resource. In some scenarios, when there are multiple policies configured against a particular pod and if each of these policies were resulting in multiple `policyEndpoint` slices individually then Network Policy agent can potentially discard firewall rules from some active `policyEndpoint` slices during a scale down/up work flows. Issue can be dependent on the sequence of events from NP controller.


jayanthvn (approved):
lgtm

PR #280, opened 2024-06-20
Configure conntrack cache table size

jayanthvn:
*Issue #, if available:* n/a

*Description of changes:* Provide a mechanism to increase conntrack cache size to match kernel conntrack table size

Note, this configuration should be made on new nodes before enabling network policy or if network policy is already enabled this would need a reload of the nodes. Dynamic update of conntrack map size would lead to traffic disruption hence we won't support it now..

```
5: lru_hash  flags 0x0
        key 20B  value 1B  max_entries 524288  memlock 50335744B
        
    --enable-ipv6=false
      --enable-network-policy=true
      --enable-cloudwatch-logs=false
      --enable-policy-event-logs=false
      --metrics-bind-addr=:8162
      --health-probe-bind-addr=:8163
      --conntrack-cache-cleanup-period=300
      --conntrack-cache-table-size=524288
      
Type : 9 ID : 5
Keysize 20 Valuesize 1 MaxEntries 524288
========================================================================================      
```

We also have a floor and ceil for the configuration -

```
{"level":"info","ts":"2024-06-20T21:52:13.365Z","logger":"setup","caller":"workspace/main.go:149","msg":"Invalid conntrack cache table size, should be between 32K and 1024K"}{"level":"error","ts":"2024-06-20T21:52:13.365Z","logger":"setup","caller":"workspace/main.go:87","msg":"Controller flags validation failed"}
```


controllers/policyendpoints_controller_test.go
```diff
@@ -748,7 +748,7 @@ func TestArePoliciesAvailableInLocalCache(t *testing.T) {
 
 		mockClient := mock_client.NewMockClient(ctrl)
 		policyEndpointReconciler, _ := NewPolicyEndpointsReconciler(mockClient, logr.New(&log.NullLogSink{}),
-			false, false, false, false, 300)
+			false, false, false, false, 300, 524288)
```
achevuru, line 751: Better to use a constant instead for the default size. Also, the default size above is different to our current [default](https://github.com/aws/aws-network-policy-agent/blob/main/pkg/ebpf/c/v4events.bpf.c#L46) in the bpf prog. I think we can leave this @ 256K.

achevuru: Ok, I see `defaultConntrackCacheTableSize ` defined and set to 256K. We can use that here.

main.go
```diff
@@ -136,3 +141,13 @@ func getLoggerWithLogLevel(logLevel string, logFilePath string) (logr.Logger, er
 	ctrlLogger := logger.New(logLevel, logFilePath)
 	return zapr.NewLogger(ctrlLogger), nil
 }
+
+// validate controller flags
+func validControllerFlags(ctrlConfig config.ControllerConfig) bool {
```
achevuru, line 146: Probably better to move it to `controller_config.go` and return an error if the validation fails and the err check above should catch it.

jayanthvn:
PR updated..

achevuru (approved):

README.md
```diff
@@ -82,6 +82,23 @@ Network Policy agent can operate in either IPv4 or IPv6 mode. Setting this flag
...
 
+#### `conntrack-cache-cleanup-period` (from v1.0.7+)
+
+Type: Integer
+
+Default: 300
+
+Network Policy agent maintains a local conntrack cache. This configuration (in seconds) will determine how fast the local conntrack cache should be cleaned up from stale/expired entries. Based on the time interval set, network policy agent checks every entry in the local conntrack cache with kernel conntrack table and determine if the entry has to be deleted.
+
+#### `conntrack-table-cache-size` (from v1.1.3+)
```
younsl, line 93: @jayanthvn There's typo.

It should be `--conntrack-cache-table-size`, not `--conntrack-table-cache-size`.

- **Reference**: https://github.com/aws/aws-network-policy-agent/pull/280/files#diff-62f7206b7bbef4728f451c0c5ce10690803f639d731ebf6f4364856645c7b58eR45

jayanthvn: Thanks, will fix this up

younsl: @jayanthvn It's okay. I already fixed it #287

PR #289, opened 2024-07-19
Rule sorting/Strict mode fix

jchen6585:
*Issue #, if available:*
We currently pass in "/32" as a parameter which will fail when converting to an int.
We strict mode is on and np agent is off, np agent will crashloop.

*Description of changes:*
Parse the "/32" we pass in

Add a check if np agent is on when validating strict mode requests


pkg/utils/utils.go
```diff
@@ -165,6 +166,7 @@ func ComputeTrieValue(l4Info []v1alpha1.Port, log logr.Logger, allowAll, denyAll
 
 	for _, l4Entry := range l4Info {
 		if startOffset >= TRIE_VALUE_LENGTH {
+			log.Error(errors.New("trie value exceeds max trie value length"), "trie value exceeds the 24 port/protocol combinations")
```
achevuru, line 169: `trie` might not be intuitive for the log msg. Instead we can use - `No.of unique port/protocol combinations supported for a single endpoint exceeded the supported maximum of 24`.

achevuru (approved):

PR #296, opened 2024-08-01
updating policy event logs with metadata

emilyhuaa:
Issue #, if available:
When users enable policy event logs, it only contains IP and Port information for Source and Destination, and users have to look up name and namespace for the pod of the given IP address. This creates inefficiency for debugging network policy issues.

Description of changes:

This PR improves the visibility by supporting the Network Policy Agent to fetch pod Kubernetes metadata through a gRPC call with a new deployment on the customer's cluster. This deployment maintains a local cache of pod and service IP address and name/namespace info from the Kubernetes API server, syncing every 10 seconds.

Manual Testing
I did some testing of each of the functions and now the policy event logs look like this:
{"level":"info","ts":"2024-07-23T23:03:50.718Z","logger":"ebpf-client","caller":"utils/utils.go:106","msg":"Flow Info: ","Src IP":"192.168.42.144","Src Name":"client-5467bc68f4-zcl6g","Src Namespace":"client-5467bc68f4-zcl6g","Src Port":34765,"Dest IP":"192.168.77.216","Dest Name":"coredns-787cb67946-blpcn","Dest Namespace":"kube-system","Dest Port":53,"Proto":"UDP","Verdict":"DENY"}


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-ARG golang_image
```
oliviassss, line 2: no need to change Dockerfile

Makefile
```diff
@@ -1,9 +1,9 @@
 
 # Image URL to use all building/pushing image targets
-IMAGE ?= amazon/aws-network-policy-agent
```
oliviassss, line 3: I assume this change is for testing purpose, please revert

pkg/ebpf/events/events.go
```diff
@@ -172,29 +171,38 @@ func capturePolicyEvents(ringbufferdata <-chan []byte, log logr.Logger, enableCl
 					continue
 				}
 
+				srcIP := utils.ConvByteToIPv6(rb.SourceIP).String()
+				srcN, srcNS := utils.GetPodMetadata(srcIP)
```
oliviassss, line 175: nit: use `srcName` instead, same for below `destName`

oliviassss: I think better to handle the error in `GetPodMetadata`, if it returns error, we should add a warning, and continue to log without name and namespace.
Maybe encapsulate all the code changes in a separate function like 
```
func logPolicyEventWithK8sMetadata(srcIP, destIP, logger) {
  1. get the metadata for srcIP, if error, add a log.Info("failed to get name and namespace for srcIP xxx')
  2. get the metadata for destIP, if error, add a log.Warning("failed to get name and namespace for destIP xxx")
  3. utils.LogFlowInfo(...)
}
```

pkg/rpc/rpc_handler.go
```diff
@@ -87,8 +94,49 @@ func (s *server) EnforceNpToPod(ctx context.Context, in *rpc.EnforceNpRequest) (
...
+			s.log.Error(err, "Failed to sync local cache with metadata cache")
+			time.Sleep(30 * time.Second)
+			continue
+		}
+
+		newCache := make(map[string]utils.Metadata)
+		for _, ipMetadata := range res.Data {
+			newCache[ipMetadata.Ip] = utils.Metadata{Name: ipMetadata.Metadata.Name, Namespace: ipMetadata.Metadata.Namespace}
+		}
+		utils.UpdateLocalCache(newCache)
```
oliviassss, line 131: maybe need to handle the case where it failed to update the local cache?

pkg/rpc/rpc_handler.go
```diff
@@ -98,7 +146,30 @@ func RunRPCHandler(policyReconciler *controllers.PolicyEndpointsReconciler) erro
...
+
+	// Connect to metadata cache service
+	serviceIP, err := utils.GetServiceIP(clientset, "metadata-cache-service", "default", rpcLog)
+	if err != nil {
+		rpcLog.Error(err, "unable to get metadata-cache-service IP")
+		os.Exit(1)
+	}
+
+	cacheAddr := serviceIP + ":50051"
+	cacheClient, err := newCacheClient(cacheAddr)
```
oliviassss, line 158: nit: no need for a `cacheAddr`, unless it's also used elsewhere

pkg/utils/utils.go
```diff
@@ -36,8 +41,103 @@ var (
...
+	LocalCache                           = make(map[string]Metadata)
+	LocalCacheMutex      sync.Mutex
 )
 
+type Metadata struct {
+	Name      string
+	Namespace string
+}
+
+// get Service IP using DNS lookup
```
oliviassss, line 53: if this piece of code is not needed, let's remove them

pkg/utils/utils.go
```diff
@@ -36,8 +41,103 @@ var (
...
+		LocalCache[ip] = metadata
+	}
+}
+
+func GetPodMetadata(ip string) (string, string) {
+	LocalCacheMutex.Lock()
+	defer LocalCacheMutex.Unlock()
+	if metadata, exists := LocalCache[ip]; exists {
+		return metadata.Name, metadata.Namespace
+	}
```
oliviassss, line 97: Maybe add an error for non-existing IPs.

pkg/ebpf/events/events.go
emilyhuaa: Right now, my logic is that if the dest name and namespace is "" then I assume it is an external service destination so should I still handle the error of if there is no name and namespace for dest IP?

pkg/rpc/rpc_handler.go
emilyhuaa: Do you mean other than logging that it failed and waiting 30 seconds to retry?

pkg/ebpf/events/events.go
```diff
@@ -207,6 +209,18 @@ func capturePolicyEvents(ringbufferdata <-chan []byte, log logr.Logger, enableCl
 	}(ringbufferdata)
 }
 
+func logPolicyEventsWithK8sMetadata(log logr.Logger, message *string, nodeName, srcIP string, srcPort int, destIP string, destPort int, protocol, verdict string) {
+	srcName, srcNS, err := utils.GetPodMetadata(srcIP)
+	if err != nil {
+		log.Info("Failed to get source name and namespace metadata", "source IP:", srcIP)
```
achevuru, line 215: Let's do away with this. We're already making an assumption that an unknown IP is an external IP to the cluster...

pkg/ebpf/events/events.go
```diff
@@ -207,6 +209,18 @@ func capturePolicyEvents(ringbufferdata <-chan []byte, log logr.Logger, enableCl
...
+	}
+	destName, destNS, err := utils.GetPodMetadata(destIP)
+	if err != nil {
+		log.Info("Failed to get destination name and namespace metadata for", "dest IP:", destIP)
```
achevuru, line 219: Same as above. Let's directly call `utils.LogFlowInfo` instead from `capturePolicyEvents`

pkg/rpc/rpc_handler.go
```diff
@@ -98,7 +146,29 @@ func RunRPCHandler(policyReconciler *controllers.PolicyEndpointsReconciler) erro
 		return errors.Wrap(err, "network policy agent: failed to listen to gRPC port")
 	}
 	grpcServer := grpc.NewServer()
-	rpc.RegisterNPBackendServer(grpcServer, &server{policyReconciler: policyReconciler, log: rpcLog})
+
+	// Connect to metadata cache service
+	serviceIP, err := utils.GetServiceIP(clientset, "metadata-cache-service", "default", rpcLog)
```
achevuru, line 151: We're making an assumption that aws-k8s-metadata service is available and we exit if we fail to find such a service on the cluster. We can't exit out of the process. We should make this optional (i.e.,) we print an enhanced log message if we find the service and if not we print the 5 tuple info of the flow (SrcIP/SrcPort/DstIP/DstPort/Protocol) like we do today.

pkg/ebpf/events/events.go
```diff
@@ -172,29 +171,32 @@ func capturePolicyEvents(ringbufferdata <-chan []byte, log logr.Logger, enableCl
...
+
 				protocol := utils.GetProtocol(int(rb.Protocol))
 				verdict := getVerdict(int(rb.Verdict))
 
-				log.Info("Flow Info:  ", "Src IP", utils.ConvByteArrayToIP(rb.SourceIP), "Src Port", rb.SourcePort,
-					"Dest IP", utils.ConvByteArrayToIP(rb.DestIP), "Dest Port", rb.DestPort,
-					"Proto", protocol, "Verdict", verdict)
-
-				message = "Node: " + nodeName + ";" + "SIP: " + utils.ConvByteArrayToIP(rb.SourceIP) + ";" + "SPORT: " + strconv.Itoa(int(rb.SourcePort)) + ";" + "DIP: " + utils.ConvByteArrayToIP(rb.DestIP) + ";" + "DPORT: " + strconv.Itoa(int(rb.DestPort)) + ";" + "PROTOCOL: " + protocol + ";" + "PolicyVerdict: " + verdict
+				logPolicyEventsWithK8sMetadata(log, &message, nodeName, srcIP, srcPort, destIP, destPort, protocol, verdict)
```
achevuru, line 199: We should call this only if we find the aws-k8s-metadata service running on the cluster. If not, we print the log in the current format and move on. If not, we will see empty values for metadata fields in the log messages and that will look odd..

pkg/rpc/rpc_handler.go
achevuru: `metadata-cache-service` --> let's make this a const and maybe add an `aws` prefix to it.

achevuru: I see that we're creating a K8S client obj just to find the service IP of metadata cache service and there is no need to do it. We can just resolve the service name (DNS resolution) and that should provide us with service IP.

main.go
```diff
@@ -83,6 +85,18 @@ func main() {
 		os.Exit(1)
 	}
 
+	k8sconfig, err := rest.InClusterConfig()
```
achevuru, line 88: I don't see a need for this. I see that we rely on this for deriving the service IP of metadata cache service and instead we can just resolve the metadata service name to get the service IP..

pkg/utils/utils.go
```diff
@@ -36,8 +41,86 @@ var (
...
+	defer LocalCacheMutex.Unlock()
+	if metadata, exists := LocalCache[ip]; exists {
+		return metadata.Name, metadata.Namespace, nil
+	}
+	return "", "", fmt.Errorf("pod metadata not found for IP %s in current cache", ip)
+}
+
+func LogFlowInfo(log logr.Logger, message *string, nodeName, srcIP, srcN, srcNS string, srcPort int, destIP, destN, destNS string, destPort int, protocol, verdict string) {
+	switch {
+	case srcN == "" && srcNS == "": // if no metadata for source IP
```
achevuru, line 87: We might've to rearrange this flow. Let's say `destN` and `destNS` is empty as well, we will end up printing "" strings for `DN` and `DNS`. Same applies to rest of the combinations below...

Also, why are we using `SHOSTIP` for Source IP field here?

emilyhuaa: If it is a host networking pod such as aws-node pods or kube-proxy pods, we agreed to just say state that the IP is a host IP with no name or namespace.

go.mod
```diff
@@ -1,52 +1,53 @@
...
 
-go 1.21.11
+go 1.22.4
 
 require (
 	github.com/aws/amazon-vpc-cni-k8s v1.18.1
 	github.com/aws/aws-ebpf-sdk-go v1.0.9
 	github.com/aws/aws-sdk-go v1.50.30
-	github.com/go-logr/logr v1.4.1
+	github.com/emilyhuaa/policyLogsEnhancement v1.0.2
```
pviossat, line 9: I'm a AWS CNI policy agent user and I was giving a look to this PR. I think adding this kind of dependency is a bad idea.

oliviassss: @pviossat, thanks for taking a look. It's a PoC project we are experimenting on, ideally the dep should be hosted under some official org, and we need to figure it out (that's why we branch out this PR). But at this point we are testing on the functionality to improve the observability of NPA event logs.

pkg/rpc/rpc_handler.go
```diff
@@ -98,7 +147,27 @@ func RunRPCHandler(policyReconciler *controllers.PolicyEndpointsReconciler) erro
...
-	rpc.RegisterNPBackendServer(grpcServer, &server{policyReconciler: policyReconciler, log: rpcLog})
+
+	// Connect to metadata cache service
+	serviceIP, _ := utils.GetServiceIP(clientset, metadataServiceName, metadataServiceNamespace)
+	cacheClient, err := newCacheClient(serviceIP + ":50051")
+	if err != nil {
+		rpcLog.Error(err, "unable to connect to aws-k8s-metadata service, continuing without")
+		utils.CacheClientConnected = false
+	} else {
+		utils.CacheClientConnected = true
```
jayanthvn, line 158: If the CacheClientConnected is `false`, why do we need to still run `syncLocalCache` ?

pkg/utils/utils.go
```diff
@@ -36,8 +41,90 @@ var (
...
+		log.Info("Flow Info: ", "Src IP", srcIP, "Src Port", srcPort,
+			"Dest IP", destIP, "Dest Name", destName, "Dest Namespace", destNS, "Dest Port", destPort, "Proto", protocol, "Verdict", verdict)
+		*message = "Node: " + nodeName + ";" + "SHOSTIP: " + srcIP + ";" + "SPORT: " + strconv.Itoa(srcPort) + ";" +
+			"DIP: " + destIP + ";" + "DN" + destName + ";" + "DNS" + destNS + ";" + "DPORT: " + strconv.Itoa(destPort) + ";" + "PROTOCOL: " + protocol + ";" + "PolicyVerdict: " + verdict
+
+	case srcName != "" && srcNS != "" && destName == "" && destNS == "": // if no metadata for dest IP (assumming it is external)
+		log.Info("Flow Info: ", "Src IP", srcIP, "Src Name", srcName, "Src Namespace", srcName, "Src Port", srcPort,
+			"Dest Ext Srv IP", destIP, "Dest Port", destPort, "Proto", protocol, "Verdict", verdict)
+		*message = "Node: " + nodeName + ";" + "SIP: " + srcIP + ";" + "SN" + srcName + ";" + "SNS" + srcName + ";" + "SPORT: " + strconv.Itoa(srcPort) + ";" +
+			"DEXTSRVIP: " + destIP + ";" + "DPORT: " + strconv.Itoa(destPort) + ";" + "PROTOCOL: " + protocol + ";" + "PolicyVerdict: " + verdict
```
jayanthvn, line 102: nit - "DEXTSRVIP" -> "DESTSRVIP"

PR #317, opened 2024-10-08
Add --ignore-loopback flag to exclude self-referential pod connections, updating cyclonus.

yash97:
*Issue #, if available:*

*Description of changes:*

This PR introduces the --ignore-loopback=true parameter to exclude test cases where a network policy blocks a pod's connection to itself. Our network policy agent attaches tc cls hooks on the host veth side, not on the container's loopback interface. Consequently, self-referential connection tests may fail, as they don't accurately reflect our agent's behavior.

This flag allows for more accurate testing by focusing on inter-pod and external communications, aligning test results with our network policy agent's actual capabilities. 

### DIFF
--- a/scripts/lib/tests.sh
+++ b/scripts/lib/tests.sh
@@ -23,11 +23,13 @@ spec:
       containers:
         - name: cyclonus
           imagePullPolicy: Always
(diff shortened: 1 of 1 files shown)


jaydeokar:
Can you check this against v6 too ? We also need to change the python script which we run to validate the results

PR #321, opened 2024-10-21
Skip v4 rules on v6 cluster and vice versa

Pavani-Panakanti:
*Issue #, if available:*

*Description of changes:*
EKS clusters are single stack and NP agent is initialized with v4 or v6 structures based on it is a v4 or v6 cluster. Applying v4 rules to v6 cluster and vice versa will result in failures while updating ebpf maps. This will result in map update failures for any other rules present in the same policy. To avoid that, skip any v4 rules in v6 cluster and vice versa


orsenthil:
Could you paste the output of the `kubectl describe networkpolicy` and `kubectl get policyendpoint` with this change?

Do we have any unit test for this change?

orsenthil (approved):
LGTM.

Pavani-Panakanti:
Added logs, networkpolicies and policyendpoints from the tests with this change to the PR. I will create a followup PR for the unit tests

Issue #258, opened 2024-04-24
CrashloopBackoff after upgrade to 1.17.1


James-Quigley:
**What happened**:
certain Pods Crashlooping after upgrading versions of aws-vpc-cni

**Attach logs**
```
2024-04-24 11:12:32.290147858 +0000 UTC Logger.check error: failed to get caller
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xaaaaea2a319c]
goroutine 99 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.0/pkg/internal/controller/controller.go:116 +0x1a4
panic({0xaaaaeb14afa0?, 0xaaaaec6eade0?})
	/root/sdk/go1.21.7/src/runtime/panic.go:914 +0x218
github.com/aws/aws-network-policy-agent/controllers.(*PolicyEndpointsReconciler).configureeBPFProbes(0x40004de000, {0xaaaaeb51dff8, 0x4000794b10}, {0x40006027d0, 0x44}, {0x4000644d40?, 0x1, 0x0?}, {0x40006bd100, 0x2, ...}, ...)
	/workspace/controllers/policyendpoints_controller.go:292 +0x34c
github.com/aws/aws-network-policy-agent/controllers.(*PolicyEndpointsReconciler).reconcilePolicyEndpoint(0x40004de000, {0xaaaaeb51dff8, 0x4000794b10}, 0x4000655520)
	/workspace/controllers/policyendpoints_controller.go:266 +0x58c
github.com/aws/aws-network-policy-agent/controllers.(*PolicyEndpointsReconciler).reconcile(0x40004de000, {0xaaaaeb51dff8, 0x4000794b10}, {{{0x4000489f80, 0x1c}, {0x400005bdc0, 0x32}}})
	/workspace/controllers/policyendpoints_controller.go:145 +0x1a4
github.com/aws/aws-network-policy-agent/controllers.(*PolicyEndpointsReconciler).Reconcile(0x40004de000, {0xaaaaeb51dff8, 0x4000794b10}, {{{0x4000489f80, 0x1c}, {0x400005bdc0, 0x32}}})
[...]

PR #323, opened 2024-10-23
Avoid panic when network policy agent is disabled

jaydeokar:
*Issue #, if available:*
Fixes -> https://github.com/aws/aws-network-policy-agent/issues/258

*Description of changes:*
When Network Policy Agent is disabled, but there are still some Policy Endpoints in the cluster, the agent fails. This was an intentional behavior to let the user know about cleaning up the stale policy endpoints in the cluster. 
With this change, the agent panic is removed and the reconciliation of policyendpoints is ignored when the agent is disabled. 
This is to support use case where a user needs to migrate from one NP implementation to another gradually.


yash97 (dismissed):
lgtm 🚀

Pavani-Panakanti (approved):

PR #344, opened 2024-12-06
Handle overlaps in except and allow CIDRs

Pavani-Panakanti:
*Issue #, if available:*
If there is an overlap of IPs in an allow rule and in an except block of other rule, we do not handle the merging 

*Description of changes:*

*Example:*

```
egress:
  - to:
      - ipBlock:
          cidr: 192.168.0.0/16
    ports:
      - port: 3306
  - to:
      - ipBlock:
          cidr: 0.0.0.0/0
          except:
            - 192.168.0.0/16
```

*We evaluate this as*

* 0.0.0.0/0 - startPort:0 endPort: 0
* except block - 192.168.0.0/16 - startPort: 0 endPort:0
* 192.168.0.0/16 - allow on port 3306, as this is part of 0.0.0.0/0 allow on all ports. Append previous ports to this new L4 info. Final L4 ports info looks as below
    * startPort: 3306, endPort: 0 - allow 
    * startPort:0, endPort: 0 - allow 
    * startPort:0, endPort:0 - deny 

The correct behavior for 192.168.0.0/16 should be allow on 3306 as explicit allow takes precedence and deny on rest all ports due to except
*New evaluation logic - all except blocks are handled at end*

* 0.0.0.0/0 - startPort:0 endPort: 0
* 192.168.0.0/16 - allow on port 3306. This is part of 0.0.0.0/0, so check if it is part of any except under this cidr. If yes do not add ports of 0.0.0.0/0 to 192.168.0.0/16. If no, add ports. In this case it is part of except so do not add it
* Handle all except at end. For every except key, check if entry is already present in cidrsMap. If yes, there is some explicit allow and everything else will be default deny so no change required. If no, we need to add deny all entry to make sure we deny all traffic and do not allow it if it is part of some other superset CIDR

*Other Changes in the PR:*

* Code refactoring to make it simpler and easy to follow
    * catch all need not be handled separately. We evaluate rules in the order of prefix length, so catch all will be handled first followed by others
* If no protocol specified, it should be allow any protocol. Changed default to any_ip_protocol from tcp
* For any_ip_protocol, check start and end ports for traffic evaluation. Before fix if it is any_protocol we allow on all ports
    * example:
        * Port:53 → This should be evaluated as allow any_protocol on port 53
        
Testing done with multiple cases of except overlaps

### DIFF
--- a/pkg/ebpf/bpf_client.go
+++ b/pkg/ebpf/bpf_client.go
@@ -915,16 +915,15 @@ func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirew
 	for _, firewallRule := range firewallRules {
 		if firewallRule.Except != nil {
 			for _, exceptCidr := range firewallRule.Except {
-				exceptFirewall := EbpfFirewallRules{
-					IPCidr: exceptCidr,
-					Except: []v1alpha1.NetworkAddress{},
-					L4Info: []v1alpha1.Port{},
+				if _, ok := cidrsMap[string(exceptCidr)]; !ok {
+					exceptFirewall := EbpfFirewallRules{
+						IPCidr: exceptCidr,
+						Except: []v1alpha1.NetworkAddress{},
+						L4Info: []v1alpha1.Port{},
+					}
+					l.addDenyAllL4Entry(&exceptFirewall)
+					cidrsMap[string(exceptCidr)] = exceptFirewall
 				}
-				if existingFirewallRuleInfo, ok := cidrsMap[string(exceptCidr)]; ok {
-					exceptFirewall.L4Info = existingFirewallRuleInfo.L4Info
-				}
-				l.addDenyAllL4Entry(&exceptFirewall)
-				cidrsMap[string(exceptCidr)] = exceptFirewall
 				l.logger.Info("Parsed Except CIDR", "IP Key: ", string(exceptCidr))
 			}
 		}
--- a/pkg/ebpf/bpf_client.go
+++ b/pkg/ebpf/bpf_client.go
@@ -897,7 +897,6 @@ func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirew
 		}
 
 		if existingFirewallRuleInfo, ok := cidrsMap[string(firewallRule.IPCidr)]; ok {
-			l.logger.Info("[DEBUG] found existing entry appending")
 			firewallRule.L4Info = append(firewallRule.L4Info, existingFirewallRuleInfo.L4Info...)
 			firewallRule.Except = append(firewallRule.Except, existingFirewallRuleInfo.Except...)
 		} else {
@@ -906,7 +905,6 @@ func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirew
 			// we use LPM TRIE map and the /m will always win out.
 			cidrL4Info = l.checkAndDeriveL4InfoFromAnyMatchingCIDRs(string(firewallRule.IPCidr), cidrsMap)
 			if len(cidrL4Info) > 0 {
-				l.logger.Info("[DEBUG] found cidrl4info from cidr matching. appending")
 				firewallRule.L4Info = append(firewallRule.L4Info, cidrL4Info...)
 			}
 		}
@@ -923,12 +921,11 @@ func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirew
 					L4Info: []v1alpha1.Port{},
 				}
 				if existingFirewallRuleInfo, ok := cidrsMap[string(exceptCidr)]; ok {
-					l.logger.Info("[DEBUG] found existing entry appending")
 					exceptFirewall.L4Info = existingFirewallRuleInfo.L4Info
 				}
 				l.addDenyAllL4Entry(&exceptFirewall)
 				cidrsMap[string(exceptCidr)] = exceptFirewall
-				l.logger.Info("Updating map with Except CIDR", "IP Key: ", string(exceptCidr))
+				l.logger.Info("Parsed Except CIDR", "IP Key: ", string(exceptCidr))
 			}
 		}
 	}
@@ -961,9 +958,8 @@ func (l *bpfClient) checkAndDeriveL4InfoFromAnyMatchingCIDRs(firewallRule string
 			continue
 		}
 		_, cidrEntry, _ := net.ParseCIDR(cidr)
-		l.logger.Info("Checking CIDR match: ", "for IP: ", firewallRule, "in CIDR: ", cidr)
 		if cidrEntry.Contains(ipToCheck.IP) {
-			l.logger.Info("Found a CIDR match: ", "for IP: ", firewallRule, "in CIDR: ", cidr)
+			l.logger.Info("Found CIDR match: ", "for IP: ", firewallRule, "in CIDR: ", cidr)
 			// If CIDR contains IP, check if it is part of any except block under CIDR. If yes, do not include cidrL4Info
 			foundInExcept := false
 			for _, except := range cidrFirewallInfo.Except {
--- a/pkg/ebpf/bpf_client.go
+++ b/pkg/ebpf/bpf_client.go
@@ -49,6 +49,7 @@ var (
 	CONNTRACK_MAP_PIN_PATH                     = "/sys/fs/bpf/globals/aws/maps/global_aws_conntrack_map"
 	POLICY_EVENTS_MAP_PIN_PATH                 = "/sys/fs/bpf/globals/aws/maps/global_policy_events"
 	CATCH_ALL_PROTOCOL         corev1.Protocol = "ANY_IP_PROTOCOL"
+	DENY_ALL_PROTOCOL          corev1.Protocol = "RESERVED_IP_PROTOCOL_NUMBER"
 	POD_VETH_PREFIX                            = "eni"
 )
 
@@ -856,10 +857,7 @@ func mergeDuplicateL4Info(ports []v1alpha1.Port) []v1alpha1.Port {
 func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirewallRules) (map[string][]byte, error) {
 
 	firewallMap := make(map[string][]byte)
-	ipCIDRs := make(map[string][]v1alpha1.Port)
-	nonHostCIDRs := make(map[string][]v1alpha1.Port)
-	isCatchAllIPEntryPresent, allowAll := false, false
-	var catchAllIPPorts []v1alpha1.Port
+	cidrsMap := make(map[string]EbpfFirewallRules)
 
 	//Traffic from the local node should always be allowed. Add NodeIP by default to map entries.
 	_, mapKey, _ := net.ParseCIDR(l.nodeIP + l.hostMask)
@@ -869,16 +867,6 @@ func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirew
 
 	//Sort the rules
 	sortFirewallRulesByPrefixLength(firewallRules, l.hostMask)
-
-	//Check and aggregate L4 Port Info for Catch All Entries.
-	catchAllIPPorts, isCatchAllIPEntryPresent, allowAll = l.checkAndDeriveCatchAllIPPorts(firewallRules)
-	if isCatchAllIPEntryPresent {
-		//Add the Catch All IP entry
-		_, mapKey, _ := net.ParseCIDR("0.0.0.0/0")
-		key := utils.ComputeTrieKey(*mapKey, l.enableIPv6)
-		value := utils.ComputeTrieValue(catchAllIPPorts, l.logger, allowAll, false)
-		firewallMap[string(key)] = value
-	}
 
 	for _, firewallRule := range firewallRules {
 		var cidrL4Info []v1alpha1.Port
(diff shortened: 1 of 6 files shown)


jayanthvn:
Can you also add a test case for this in our integration suite?

pkg/ebpf/bpf_client.go
```diff
@@ -856,10 +857,7 @@ func mergeDuplicateL4Info(ports []v1alpha1.Port) []v1alpha1.Port {
 func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirewallRules) (map[string][]byte, error) {
```
orsenthil, line 857: As an enhancement, could you document the behavior of this function as function doc string itself. We pack a lot inside the function here.

pkg/ebpf/c/tc.v4egress.bpf.c
```diff
@@ -201,7 +201,8 @@ int handle_egress(struct __sk_buff *skb)
 				return BPF_DROP;
 			}
 
-			if ((trie_val->protocol == ANY_IP_PROTOCOL) || (trie_val->protocol == ip->protocol &&
+			if ((trie_val->protocol == ANY_IP_PROTOCOL && ((trie_val->start_port == ANY_PORT) || (l4_dst_port == trie_val->start_port) ||
```
orsenthil, line 204: Enhancement Request: We can format this and explain the behavior of in the comment.
a) match any problem and ensure either destination port matches or falls within the range.

pkg/ebpf/bpf_client_test.go
```diff
@@ -175,31 +175,18 @@ func TestBpfClient_IsEBPFProbeAttached(t *testing.T) {
 	}
 }
 
-func TestBpfClient_CheckAndDeriveCatchAllIPPorts(t *testing.T) {
+func TestBpfClient_CheckAndDeriveL4InfoFromAnyMatchingCIDRs(t *testing.T) {
```
orsenthil, line 178: We need more test cases here that will cover for the scenarios described in the description of the PR are handled by the test case.

```
192.168.0.0/16 - allow on port 3306. This is part of 0.0.0.0/0, so check if it is part of any except under this cidr. If yes do not add ports of 0.0.0.0/0 to 192.168.0.0/16. If no, add ports. 

Handle all except at end. For every except key, check if entry is already present in cidrsMap. If yes, there is some explicit allow and everything else will be default deny so no change required. If no, we need to add deny all entry to make sure we deny all traffic and do not allow it if it is part of some other superset CIDR
```

The code change and the logic looks good to me. As noted in the review of the design doc, I wanted to see if there is any simpler way, but handling all except at the end looks alright.

Coverage with test case can help us push this PR to finish line here. Thank you.

pkg/ebpf/bpf_client.go
```diff
@@ -1003,13 +1004,31 @@ func mergeDuplicateL4Info(ports []v1alpha1.Port) []v1alpha1.Port {
 	return result
 }
 
+// computeMapEntriesFromEndpointRules generates a map of IP prefix keys to encoded L4 rules that will
+// be used to update ebpf maps
+//
+// How it works:
+//   1. A default allow-all entry is added for the node IP to ensure local node traffic is always permitted.
+//   2. The list of firewall rules is sorted by prefix length in descending order. This is crucial for
```
viveksb007, line 1012: I think the sorting happens in ascending order of prefix length. I was trying to understand this while refactoring and also wrote a test for this sorting fn https://github.com/aws/aws-network-policy-agent/pull/419/files#diff-36a28a4029b0030c4806fc3502df4be1c44c2f35e54196df5cdf7c93f3448d71

pkg/ebpf/bpf_client.go
```diff
@@ -1051,106 +1060,89 @@ func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirew
...
-			firewallKey := utils.ComputeTrieKey(*firewallMapKey, l.enableIPv6)
-
-			if len(firewallRule.L4Info) != 0 {
-				mergedL4Info := mergeDuplicateL4Info(firewallRule.L4Info)
-				firewallRule.L4Info = mergedL4Info
+		// If no L4 specified add catch all entry
+		if len(firewallRule.L4Info) == 0 {
+			log().Debugf("No L4 specified. Add Catch all entry CIDR: %s", string(firewallRule.IPCidr))
+			l.addCatchAllL4Entry(&firewallRule)
+			log().Debugf("Total L4 entries count: %d", len(firewallRule.L4Info))
```
haouc, line 1067: nit: I think `L4Info`'s len is 0 if getting here.

pkg/ebpf/bpf_client.go
Pavani-Panakanti: Yeah its ascending order. updated that comment

pkg/ebpf/bpf_client.go
```diff
@@ -1051,106 +1060,89 @@ func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirew
...
+			// If found, we need to include the port and protocol combination against the current entry as well since
+			// we use LPM TRIE map and the /m will always win out.
+			cidrL4Info = l.checkAndDeriveL4InfoFromAnyMatchingCIDRs(string(firewallRule.IPCidr), cidrsMap)
+			if len(cidrL4Info) > 0 {
+				firewallRule.L4Info = append(firewallRule.L4Info, cidrL4Info...)
 			}
-			firewallValue := utils.ComputeTrieValue(firewallRule.L4Info, allowAll, false)
-			firewallMap[string(firewallKey)] = firewallValue
 		}
+		cidrsMap[string(firewallRule.IPCidr)] = firewallRule
```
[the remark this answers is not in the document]
Pavani-Panakanti: Yes CRD won't allow duplicates

pkg/ebpf/bpf_client.go
```diff
@@ -1051,106 +1060,89 @@ func (l *bpfClient) computeMapEntriesFromEndpointRules(firewallRules []EbpfFirew
...
-				firewallRule.L4Info = mergedL4Info
+		// If no L4 specified add catch all entry
+		if len(firewallRule.L4Info) == 0 {
+			log().Debugf("No L4 specified. Add Catch all entry CIDR: %s", string(firewallRule.IPCidr))
+			l.addCatchAllL4Entry(&firewallRule)
+			log().Debugf("Total L4 entries count: %d", len(firewallRule.L4Info))
+		}
 
+		if existingFirewallRuleInfo, ok := cidrsMap[string(firewallRule.IPCidr)]; ok {
+			firewallRule.L4Info = append(firewallRule.L4Info, existingFirewallRuleInfo.L4Info...)
```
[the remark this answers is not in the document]
Pavani-Panakanti: Yeah CRD checks ensure that they are not empty

pkg/ebpf/c/tc.v4egress.bpf.c
```diff
@@ -111,10 +111,24 @@ static inline int evaluateByLookUp(struct keystruct trie_key, struct conntrack_k
...
+		//    - Then match if:
+		//        - start_port is ANY_PORT → rule applies to all ports
+		//        - OR l4_dst_port is exactly the start_port
+		//        - OR l4_dst_port falls within the inclusive [start_port, end_port] range
+		//
+		// 2. Specific Protocol Match:
+		//    - If trie_val->protocol matches the packet's IP protocol (e.g., TCP or UDP),
+		//    - Then apply the same port match logic as above.
+
+		if ((trie_val->protocol == ANY_IP_PROTOCOL &&
```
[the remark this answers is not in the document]
Pavani-Panakanti: Yeah, agreed the check is a bit complex, but it's generally discouraged to break it into multiple functions in BPF kernel code. Due to verifier limitations and performance constraints, inlining logic is often preferred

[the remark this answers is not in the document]
Pavani-Panakanti: Yeah, agreed the check is a bit complex, but it's generally discouraged to break it into multiple functions in BPF kernel code. Due to verifier limitations and performance constraints, inlining logic is often preferred

pkg/ebpf/c/tc.v6ingress.bpf.c
```diff
@@ -115,9 +115,22 @@ static inline int evaluateByLookUp(struct keystruct trie_key, struct conntrack_k
...
+		//
+		// 2. Specific Protocol Match:
+		//    - If trie_val->protocol matches the packet's IP protocol (e.g., TCP or UDP),
+		//    - Then apply the same port match logic as above.
+		if ((trie_val->protocol == ANY_IP_PROTOCOL &&
+			((trie_val->start_port == ANY_PORT) || (l4_dst_port == trie_val->start_port) ||
+			(l4_dst_port > trie_val->start_port && l4_dst_port <= trie_val->end_port))) || 
+			(trie_val->protocol == ip->nexthdr &&
+			((trie_val->start_port == ANY_PORT) || (l4_dst_port == trie_val->start_port) ||
+			(l4_dst_port > trie_val->start_port && l4_dst_port <= trie_val->end_port)))) {
```
haouc, line 133: Same question here, should this be inclusive as well?

pkg/ebpf/c/tc.v6egress.bpf.c
```diff
@@ -113,9 +113,22 @@ static inline int evaluateByLookUp(struct keystruct trie_key, struct conntrack_k
...
+		//        - start_port is ANY_PORT → rule applies to all ports
+		//        - OR l4_dst_port is exactly the start_port
+		//        - OR l4_dst_port falls within the inclusive [start_port, end_port] range
+		//
+		// 2. Specific Protocol Match:
+		//    - If trie_val->protocol matches the packet's IP protocol (e.g., TCP or UDP),
+		//    - Then apply the same port match logic as above.
+		if ((trie_val->protocol == ANY_IP_PROTOCOL &&
+			((trie_val->start_port == ANY_PORT) || (l4_dst_port == trie_val->start_port) ||
+			(l4_dst_port > trie_val->start_port && l4_dst_port <= trie_val->end_port))) ||
```
haouc, line 128: Should this be inclusive as comment "l4_dst_port falls within the inclusive [start_port, end_port] range"?

pkg/ebpf/c/tc.v4ingress.bpf.c
```diff
@@ -112,9 +112,22 @@ static inline int evaluateByLookUp(struct keystruct trie_key, struct conntrack_k
...
+		//        - start_port is ANY_PORT → rule applies to all ports
+		//        - OR l4_dst_port is exactly the start_port
+		//        - OR l4_dst_port falls within the inclusive [start_port, end_port] range
+		//
+		// 2. Specific Protocol Match:
+		//    - If trie_val->protocol matches the packet's IP protocol (e.g., TCP or UDP),
+		//    - Then apply the same port match logic as above.
+		if ((trie_val->protocol == ANY_IP_PROTOCOL && 
+			((trie_val->start_port == ANY_PORT) || (l4_dst_port == trie_val->start_port) ||
+			(l4_dst_port > trie_val->start_port && l4_dst_port <= trie_val->end_port))) ||
```
haouc, line 127: Should be be inclusive as commented?

pkg/ebpf/c/tc.v4egress.bpf.c
```diff
@@ -111,10 +111,24 @@ static inline int evaluateByLookUp(struct keystruct trie_key, struct conntrack_k
...
+		// 2. Specific Protocol Match:
+		//    - If trie_val->protocol matches the packet's IP protocol (e.g., TCP or UDP),
+		//    - Then apply the same port match logic as above.
+
+		if ((trie_val->protocol == ANY_IP_PROTOCOL && 
+			((trie_val->start_port == ANY_PORT) || (l4_dst_port == trie_val->start_port) ||
+			(l4_dst_port > trie_val->start_port && l4_dst_port <= trie_val->end_port))) || 
+			(trie_val->protocol == ip->protocol &&
+			((trie_val->start_port == ANY_PORT) || (l4_dst_port == trie_val->start_port) ||
+			(l4_dst_port > trie_val->start_port && l4_dst_port <= trie_val->end_port)))) {
```
haouc, line 131: Should this be inclusive as commented "l4_dst_port falls within the inclusive [start_port, end_port] range"?

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -90,94 +107,58 @@ func (f *FirewallRuleProcessor) ComputeMapEntriesFromEndpointRules(firewallRules
...
-			firewallKey := utils.ComputeTrieKey(*firewallMapKey, f.enableIPv6)
-
-			if len(firewallRule.L4Info) != 0 {
-				mergedL4Info := mergeDuplicateL4Info(firewallRule.L4Info)
-				firewallRule.L4Info = mergedL4Info
+		// If no L4 specified add catch all entry
+		if len(firewallRule.L4Info) == 0 {
+			log().Debugf("No L4 specified. Add Catch all entry CIDR: %s", string(firewallRule.IPCidr))
+			addCatchAllL4Entry(&firewallRule)
+			log().Debugf("Total L4 entries count: %d", len(firewallRule.L4Info))
```
haouc, line 114: nit: looks like two logs can be combined since they are executed under the same condition.

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -90,94 +107,58 @@ func (f *FirewallRuleProcessor) ComputeMapEntriesFromEndpointRules(firewallRules
...
+		if _, ok := cidrsMap[exceptCidr]; !ok {
+			exceptFirewall := EbpfFirewallRules{
+				IPCidr: v1alpha1.NetworkAddress(exceptCidr),
+				Except: []v1alpha1.NetworkAddress{},
+				L4Info: []v1alpha1.Port{},
 			}
+			addDenyAllL4Entry(&exceptFirewall)
+			cidrsMap[exceptCidr] = exceptFirewall
 		}
+		log().Debugf("Parsed Except CIDR: %s", exceptCidr)
```
haouc, line 143: Could this log be misleading if the "if" condition doesn't satisfy?

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -90,94 +107,58 @@ func (f *FirewallRuleProcessor) ComputeMapEntriesFromEndpointRules(firewallRules
...
-}
+	for key, value := range cidrsMap {
+		log().Infof("Updating Map with IP Key: %s", string(key))
+		_, firewallMapKey, _ := net.ParseCIDR(string(key))
+		// Key format: Prefix length (4 bytes) followed by 4/16byte IP address
+		firewallKey := utils.ComputeTrieKey(*firewallMapKey, f.enableIPv6)
+
+		if len(value.L4Info) != 0 {
+			mergedL4Info := mergeDuplicateL4Info(value.L4Info)
+			value.L4Info = mergedL4Info
```
haouc, line 154: Looks like 
```
value.L4Info = mergeDuplicateL4Info(value.L4Info)
```

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -90,94 +107,58 @@ func (f *FirewallRuleProcessor) ComputeMapEntriesFromEndpointRules(firewallRules
...
-		if utils.IsCatchAllIPEntry(string(firewallRule.IPCidr)) {
-			catchAllL4Info = append(catchAllL4Info, firewallRule.L4Info...)
-			isCatchAllIPEntryPresent = true
-			if len(firewallRule.L4Info) == 0 {
-				//All ports and protocols
-				allowAllPortAndProtocols = true
-			}
 		}
+		firewallValue := utils.ComputeTrieValue(value.L4Info, false, false)
+		firewallMap[string(firewallKey)] = firewallValue
```
haouc, line 158: same here?

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -90,94 +107,58 @@ func (f *FirewallRuleProcessor) ComputeMapEntriesFromEndpointRules(firewallRules
...
 	}
-	log().Debugf("Total L4 entry count for catch all entry: count: %d", len(catchAllL4Info))
-	return catchAllL4Info, isCatchAllIPEntryPresent, allowAllPortAndProtocols
+
+	return firewallMap, nil
```
haouc, line 161: I don't think I saw somewhere else returning (map, err). If we never return error, do we need a Error type in return?

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -212,16 +193,38 @@ func addCatchAllL4Entry(firewallRule *EbpfFirewallRules) {
 	firewallRule.L4Info = append(firewallRule.L4Info, catchAllL4Entry)
 }
 
+func addDenyAllL4Entry(firewallRule *EbpfFirewallRules) {
+	denyAllL4Entry := v1alpha1.Port{
+		Protocol: &DENY_ALL_PROTOCOL,
+	}
+	firewallRule.L4Info = append(firewallRule.L4Info, denyAllL4Entry)
```
haouc, line 200: nit: combine both lines to one line?

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -212,16 +193,38 @@ func addCatchAllL4Entry(firewallRule *EbpfFirewallRules) {
...
 func checkAndDeriveL4InfoFromAnyMatchingCIDRs(firewallRule string,
-	nonHostCIDRs map[string][]v1alpha1.Port) []v1alpha1.Port {
+	cidrsMap map[string]EbpfFirewallRules) []v1alpha1.Port {
 	var matchingCIDRL4Info []v1alpha1.Port
 
 	_, ipToCheck, _ := net.ParseCIDR(firewallRule)
-	for nonHostCIDR, l4Info := range nonHostCIDRs {
-		_, cidrEntry, _ := net.ParseCIDR(nonHostCIDR)
+	for cidr, cidrFirewallInfo := range cidrsMap {
+		if !utils.IsNonHostCIDR(cidr) {
```
haouc, line 209: nit: should the utils helper func named "IsHostCIDR" instead? :)

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -212,16 +193,38 @@ func addCatchAllL4Entry(firewallRule *EbpfFirewallRules) {
...
+			continue
+		}
+		_, cidrEntry, _ := net.ParseCIDR(cidr)
```
haouc, line 212: Do we need check parsing error to avoid / catch unexpected failures in following logics?

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -212,16 +193,38 @@ func addCatchAllL4Entry(firewallRule *EbpfFirewallRules) {
...
 		if cidrEntry.Contains(ipToCheck.IP) {
-			log().Debugf("Found a CIDR match for IP: %s in CIDR %s ", firewallRule, nonHostCIDR)
-			matchingCIDRL4Info = append(matchingCIDRL4Info, l4Info...)
+			log().Debugf("Found CIDR match or IP: %s in CIDR: %s", firewallRule, cidr)
+			// If CIDR contains IP, check if it is part of any except block under CIDR. If yes, do not include cidrL4Info
+			foundInExcept := false
+			for _, except := range cidrFirewallInfo.Except {
+				_, exceptEntry, _ := net.ParseCIDR(string(except))
```
haouc, line 218: Same here, do we need check parsing error?

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -212,16 +193,38 @@ func addCatchAllL4Entry(firewallRule *EbpfFirewallRules) {
...
+				if exceptEntry.Contains(ipToCheck.IP) {
+					foundInExcept = true
+					log().Debugf("Found IP: %s in except block %s of CIDR %s. Skipping CIDR match", firewallRule, string(except), cidr)
+					break
```
haouc, line 222: Since we break out for loop, I think we don't need the boolean variable "foundInExcept"?

pkg/utils/utils.go
```diff
@@ -331,7 +328,7 @@ func IsNodeIP(nodeIP string, ipCidr string) bool {
 func IsNonHostCIDR(ipAddr string) bool {
 	ipSplit := strings.Split(ipAddr, "/")
 	//Ignore Catch All IP entry as well
-	if ipSplit[1] != "32" && ipSplit[1] != "128" && ipSplit[1] != "0" {
+	if ipSplit[1] != "32" && ipSplit[1] != "128" {
```
haouc, line 331: Why do we remove `ipSplit[1] != "0"` check?

pkg/fwruleprocessor/fw_rule_processor.go
Pavani-Panakanti: updated code with a small optimization which removed this

pkg/fwruleprocessor/fw_rule_processor.go
Pavani-Panakanti: we still use it after we exit out of for loop

pkg/fwruleprocessor/fw_rule_processor.go
Pavani-Panakanti: Agree. Most of this func was not modified as part of this PR but earlier code was showing up in diff due to refactor. I think we should add some error handling. Rather than removing error in return we can add error handling in follow up PR's and use the return error

pkg/fwruleprocessor/fw_rule_processor.go
Pavani-Panakanti: updated

pkg/fwruleprocessor/fw_rule_processor.go
Pavani-Panakanti: updated

pkg/fwruleprocessor/fw_rule_processor.go
Pavani-Panakanti: makes sense. Second log does not seem to be adding any value. Removed it

Issue #345, opened 2024-12-06
Return traffic can be denied for a short duration once the policies are reconciled on a new pod 


Pavani-Panakanti:
**What happened**:
In standard mode, we do a default allow at pod startup and all traffic is allowed before policies are reconciled. It takes 1-2secs for the policies to be reconciled on the new pod. Once the network policy reconciliation happens, we start tracking the flows in conntrack table. For return traffic we check if entry is present in conntrack table and allow it accordingly. For traffic which exited the pod before network policies were applied and return traffic came after policies were applied, the return traffic will be denied as entry is not tracked in conntrack table

As a mitigation, 2-5secs delay can be added at the pod startup using init container. As a result, traffic will start going out of the pod only after network policies were applied and there will be no denies in the return traffic

We are actively working on fixing this issue, so that cx can use standard mode without the need to add sleep at pod startup. Fix for this issue can be tracked here 

Please note that this issue happens only with standard mode and not in strict mode

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- CNI Version
- Network Policy Agent Version
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):

youwalther65:
NetPol are always asynchronously reconciled. What makes standard mode so special here compared to strict mode? Can you please elaborate on the technical side a bit.

Pavani-Panakanti:
@youwalther65 In strict mode, we do default deny before policies are applied on the first pod, so no egress traffic goes out of the pod before policies were applied. So above issue will not happen where response packet will be denied as entry is missing in conntrack table for traffic that egressed out of pod before applying policies

Pavani-Panakanti:
@m00lecule We are looking into improving the user experience for strict mode. This is something we are prioritizing. We will provide more details soon

janavenkat:
Syncing here as well https://github.com/aws/amazon-vpc-cni-k8s/pull/3206#issuecomment-2703265280

Pavani-Panakanti:
Looking into this. Will add an update soon

Monska85:
Are there any news about this topic?

Pavani-Panakanti:
[...]

PR #361, opened 2025-02-03
Fix standard mode return packet drop at pod startup

Pavani-Panakanti:
*Issue #, if available:*
https://github.com/aws/aws-network-policy-agent/issues/345

*Description of changes:*
* Probes will be attached to pod at pod startup for both standard and strict mode 
* Probes will be deleted when pod gets deleted
* Added a new bpf map pod_state_map which stores the current state of all pods in pod identifier - DEFAULT_ALLOW, DEFAULT_DENY, POLICIES_APPLIED
* At pod startup pod_state will be in default allow for standard mode and default deny for strict mode
* Each conntrack entry will have state against - This is pod state when conntrack entry was added
* Once the policies are reconciled state will be flipped to POLICIES_APPLIED after updating bpf maps with new rules 
* Response packets will be accepted as we will have outgoing packets tracked in conntrack table. For any new connections, we will see if pod state changed from the time the conntrack entry was added and delete it if it no longer should be allowed with new rules

*Testing*
Following all cases were tested
* connection packets went out in default allow and response came after policies applied - egress and ingress side response packets for v4 and v6
* standard mode expected behavior - v4 and v6
* strict mode expected behavior - v4 and v6
* upgrades, downgrades, enable NP, disable NP
* clean up of policies, pods, caches
* Added unit tests


jayanthvn:
Need to fix the build and UTs.

controllers/policyendpoints_controller.go
```diff
@@ -58,6 +59,9 @@ var (
 		[]string{"name", "namespace"},
 	)
 	prometheusRegistered = false
+	POLICIES_APPLIED     = 0
```
jayanthvn, line 62: You can consider making this a const. Something like this - https://github.com/aws/aws-ebpf-sdk-go/blob/135349feeb1afbdce533ead8160813277d4d7387/pkg/constants/constants.go#L6-L10

controllers/policyendpoints_controller.go
```diff
@@ -208,31 +217,23 @@ func (r *PolicyEndpointsReconciler) cleanUpPolicyEndpoint(ctx context.Context, r
...
-				r.log.Info("isProgFdShared", "Found shared egress progFD for target: ", targetPodName, "progFD: ", targetProgFD)
-			}
+	if currentList, ok := r.ebpfClient.GetIngressProgToPodsMap().Load(targetProgFD); ok {
+		podsList, ok := currentList.(map[string]struct{})
+		if ok && len(podsList) > 1 {
+			foundSharedIngress = true
+			r.log.Info("isProgFdShared", "Found shared ingress progFD for target: ", targetPodName, "progFD: ", targetProgFD)
 		}
 	}
-	return foundSharedIngress || foundSharedEgress
```
jayanthvn, line 235: We need to check even egress right? egress prog might be shared by another pod - https://github.com/aws/aws-network-policy-agent/blob/bca55c5ff3d437afd30919023d39d4d744ff0ff3/controllers/policyendpoints_controller.go#L253

controllers/policyendpoints_controller.go
```diff
@@ -241,21 +242,14 @@ func (r *PolicyEndpointsReconciler) updatePolicyEnforcementStatusForPods(ctx con
...
-	//    corresponding BPF maps. We will also clean up eBPF pin paths under BPF FS.
+	// 3. If there are no more active policies against this pod, we update pod_state to default deny/allow
 	for _, targetPod := range targetPods {
 		r.log.Info("Updating Pod: ", "Name: ", targetPod.Name, "Namespace: ", targetPod.Namespace)
 
-		deletePinPath := true
 		podIdentifier := utils.GetPodIdentifier(targetPod.Name, targetPod.Namespace, r.log)
 		r.log.Info("Derived ", "Pod identifier to check if update is needed : ", podIdentifier)
-		// check if ebpf progs are being shared by any other pods
-		if progFdShared := r.isProgFdShared(ctx, targetPod.Name, targetPod.Namespace); progFdShared {
```
jayanthvn, line 253: Can you test pod churn and if there are any leaked FDs or pinpaths?

controllers/policyendpoints_controller.go
```diff
@@ -375,13 +369,22 @@ func (r *PolicyEndpointsReconciler) cleanupeBPFProbes(ctx context.Context, targe
...
 
-		// We only detach probes if there are no policyendpoint resources on both the
-		// directions
+		// We update pod_state to default allow/deny if there are no other policies applied
 		if noActiveIngressPolicies && noActiveEgressPolicies {
-			err = r.ebpfClient.DetacheBPFProbes(targetPod, noActiveIngressPolicies, noActiveEgressPolicies, deletePinPath)
-			if err != nil {
-				r.log.Info("PolicyEndpoint cleanup unsuccessful", "Name: ", policyEndpoint)
-				return err
+			if utils.IsStrictMode(r.networkPolicyMode) {
```
jayanthvn, line 374: Refactor, something like - 

```
mode := default_allow
if (strict) {
   mode = default_deny
}

r. GeteBPF...(podIdentifier, mode, true, true)
```

controllers/policyendpoints_controller.go
```diff
@@ -625,10 +628,6 @@ func (r *PolicyEndpointsReconciler) getPodListToBeCleanedUp(oldPodSet []types.Na
 			r.log.Info("Pod to cleanup: ", "name: ", oldPod.Name, "namespace: ", oldPod.Namespace)
 			podsToBeCleanedUp = append(podsToBeCleanedUp, oldPod)
 			// When pod is deleted and this is not the last pod in podIdentifier, we are just updating the podName and progFD local caches
-		} else if !activePod {
```
jayanthvn, line 628: Nit: Need to update the comments..

But this section was added for the race condition issue - https://github.com/aws/aws-network-policy-agent/pull/306#issuecomment-2374757801 due to out of order updates from CNI/controller and there is more than one replica on the same node then don't we need to update the local cache?

pkg/ebpf/bpf_client.go
```diff
@@ -294,15 +305,8 @@ type bpfClient struct {
...
-
-type Event_t struct {
-	SourceIP   uint32
-	SourcePort uint32
-	DestIP     uint32
-	DestPort   uint32
-	Protocol   uint32
-	Verdict    uint32
+	// Stores podIdentifier to deletepod lock mapping
+	DeletePodLock *sync.Map
```
jayanthvn, line 309: nit: Maybe podIdenfierLock or something else?

controllers/policyendpoints_controller.go
jayanthvn: As discussed, lets add the function back

controllers/policyendpoints_controller.go
```diff
@@ -241,21 +242,14 @@ func (r *PolicyEndpointsReconciler) updatePolicyEnforcementStatusForPods(ctx con
 	// 1. If the pods are already deleted, we move on.
 	// 2. If the pods have another policy or policies active against them, we update the maps to purge the entries
 	//    introduced by the current policy.
-	// 3. If there are no more active policies against this pod. Detach and delete the probes and
-	//    corresponding BPF maps. We will also clean up eBPF pin paths under BPF FS.
+	// 3. If there are no more active policies against this pod, we update pod_state to default deny/allow
 	for _, targetPod := range targetPods {
 		r.log.Info("Updating Pod: ", "Name: ", targetPod.Name, "Namespace: ", targetPod.Namespace)
 
-		deletePinPath := true
```
jayanthvn, line 249: I see, we update to default all or deny in cleanup but where do we delete the old pin path. I.e, if pod has a NP and we delete the NP or update NP to remove the pod selector. Will it land in `getPodListToBeCleanedUp `?

pkg/ebpf/bpf_client.go
```diff
@@ -746,6 +708,55 @@ func (l *bpfClient) UpdateEbpfMaps(podIdentifier string, ingressFirewallRules []
...
+	var ingressProgFD, egressProgFD int
+	var mapToUpdate goebpfmaps.BpfMap
+	start := time.Now()
+	value, ok := l.policyEndpointeBPFContext.Load(podIdentifier)
+
+	if ok {
+		peBPFContext := value.(BPFContext)
+		ingressProgInfo := peBPFContext.ingressPgmInfo
+		egressProgInfo := peBPFContext.egressPgmInfo
+		key := uint32(0)                        // pod_state_map key
```
jayanthvn, line 730: Nit: Consider having a constant

pkg/ebpf/c/tc.v4ingress.bpf.c
```diff
@@ -182,40 +270,20 @@ int handle_ingress(struct __sk_buff *skb)
...
 		}
 
-		//Check if it's in the allowed list
-		trie_val = bpf_map_lookup_elem(&ingress_map, &trie_key);
-		if (trie_val == NULL) {
-			evt.verdict = 0;		    
+		if (pst->state == DEFAULT_ALLOW) {
+			struct conntrack_value new_flow_val = {};
+			new_flow_val.val = CT_VAL_DEFAULT_ALLOW;
+			bpf_map_update_elem(&aws_conntrack_map, &flow_key, &new_flow_val, 0); // 0 - BPF_ANY
```
jayanthvn, line 280: Do test out events generation if you haven't already

pkg/rpc/rpc_handler.go
```diff
@@ -95,6 +125,44 @@ func (s *server) EnforceNpToPod(ctx context.Context, in *rpc.EnforceNpRequest) (
...
+	s.log.Info("Received Delete Network Policy Request for Pod", "Name", in.K8S_POD_NAME, "Namespace", in.K8S_POD_NAMESPACE)
+	var err error
+	podIdentifier := utils.GetPodIdentifier(in.K8S_POD_NAME, in.K8S_POD_NAMESPACE, s.log)
+
+	value, _ := s.policyReconciler.GeteBPFClient().GetDeletePodLockMap().LoadOrStore(podIdentifier, &sync.Mutex{})
+	deletePodLock := value.(*sync.Mutex)
+	deletePodLock.Lock()
+	s.log.Info("Got the deletePodLock for", "Pod: ", in.K8S_POD_NAME, " Namespace: ", in.K8S_POD_NAMESPACE, " PodIdentifier: ", podIdentifier)
+
+	isProgFdShared, err := s.policyReconciler.IsProgFdShared(in.K8S_POD_NAME, in.K8S_POD_NAMESPACE)
```
jayanthvn, line 147: ok, this answers one of my previous comments. So we capture the delete pod event and cleanup bpf progs and maps..

controllers/policyendpoints_controller.go
Pavani-Panakanti: updated

controllers/policyendpoints_controller.go
Pavani-Panakanti: We always attach both ingress and egress probes. So entry should either be present in both caches or not present in both. Added egress check too anyways. It will be more readable and need not have the above context

controllers/policyendpoints_controller.go
Pavani-Panakanti: updated. looks cleaner

pkg/ebpf/bpf_client.go
Pavani-Panakanti: updated

pkg/ebpf/c/tc.v4ingress.bpf.c
Pavani-Panakanti: Tested events generation as part of test cases

controllers/policyendpoints_controller.go
Pavani-Panakanti: Pods will have probes attached till its deleted because policies can be applied again and we don't want to run into the return packet issue again. Pod will start with default allow and if any policies applied it will be in policies_applied state. If policies are removed we move state to default_allow/deny. Probes deletion, caches cleanup and pinpath cleanup will all happen in pod deletion flow

controllers/policyendpoints_controller.go
Pavani-Panakanti: Updated comments 

That race condition would never happen after changes because we don't delete pinpaths anywhere in reconcile path. That happens only when pod is deleted. 

Pod1 was deleted. Pod2 was created. 
Removing pod1 from PE reconcile request came first -> we add this to cleanuppodlist -> no active policies present, state to default allow/deny
Adding Pod2 to PE reconcile request -> we updates rules, pod_state to policies_applied

programs and pinpath deletion happens only when both pods are deleted from CNI

jayanthvn:
lgtm, but lets tests the cases we discussed offline. Significant amount of code changes are going in..

jayanthvn (approved):

Issue #362, opened 2025-02-04
aws-eks-nodeagent restart in v1.19.2


jmgalvez:
**What happened**:

In AWS VPC CNI EKS addon version `v1.18.5-eksbuild.1`, _aws-node_ pod restarted due to a panic in _aws-eks-nodeagent_ container. We have upgraded to the very latest version, which is `v1.19.2-eksbuild.1`, but the issue stills persist. The issue reported here is based on the latest version.

**Attach logs**

```
{"level":"info","ts":"2025-01-22T08:53:09.976Z","caller":"metrics/metrics.go:23","msg":"Serving metrics on ","port":61680}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x559f468360fc]

goroutine 84 [running]:
github.com/aws/aws-network-policy-agent/pkg/ebpf/events.publishDataToCloudwatch({0x0, 0x0, 0x0}, {0xc000bdcbe0, 0x93}, {{0x559f4877ae38?, 0xc00078e3f0?}, 0xc?})
	/workspace/pkg/ebpf/events/events.go:150 +0x43c
github.com/aws/aws-network-policy-agent/pkg/ebpf/events.capturePolicyEvents.func1(0xc000230420)
	/workspace/pkg/ebpf/events/events.go:201 +0xe54
created by github.com/aws/aws-network-policy-agent/pkg/ebpf/events.capturePolicyEvents in goroutine 1
	/workspace/pkg/ebpf/events/events.go:162 +0x127
```

Yesterday we saw a new restart at 14.46:

```
  aws-eks-nodeagent:
    Container ID:  containerd://156d74ac7d633027ecdaaf1edbc2a50124c762ee6ff86512c8801819105883bd
    Image:         602401143452.dkr.ecr.eu-central-1.amazonaws.com/amazon/aws-network-policy-agent:v1.1.6-eksbuild.1
    Image ID:      602401143452.dkr.ecr.eu-central-1.amazonaws.com/amazon/aws-network-policy-agent@sha256:8f53fe281da5074976526a28a13422134a23aeab33199c0947fa61b34086eac3
    Port:          <none>
    Host Port:     <none>
    Args:
      --enable-ipv6=false
      --enable-network-policy=true
      --enable-cloudwatch-logs=true
      --enable-policy-event-logs=true
      --log-file=/var/log/aws-routed-eni/network-policy-agent.log
[...]

PR #363, opened 2025-02-06
Fix error handling while publishing events to cloudwatch

Pavani-Panakanti:
*Issue #, if available:*
https://github.com/aws/aws-network-policy-agent/issues/362

*Description of changes:*
When resp is maformed or we get partial response it will crash as we try to reference nil pointer. 
Check resp only if err is nil and also check if resp.nextSequenceToken is not nil 


jaydeokar (approved):
lgtm

PR #365, opened 2025-02-11
Attach new probes at start if probe update required

Pavani-Panakanti:
*Issue #, if available:*
If there is a change is bpf binary update, today we just delete the existing filters. With this we will put strict mode pods into default allow for a very short time. For standard mode, we might run into return packet drop issue again. To resolve these two issues, if there is a bpf binary update we attach new probes with new binary

*Description of changes:*
If there is no bpf binary change --> recover existing programs and maps and store them. Same probes and filters will be applied for the traffic 
If there is a bpf binary change --> Get all loaded bpf programs and maps. Get all links which have filters attached. Match filter progId to pinpath. Delete the existing filters. Create new bpf programs and maps with new binary and attach them to the links

Pods which had probes attached before upgrade will have probes attached even after upgrade too with this change. If no update required --. same probes. If update required -- clean existing and attach new with new binaries

*Testing*
Tested following scenarios 
- Programs are being loaded and new probes are being attached to all pods as expected 
- Call to ipamd works as expected - no issues
- This reattach gets triggered only when there is binary update 
- Incase of any failure we wait for reconcile request to attach probes


pkg/ebpf/bpf_client.go
```diff
@@ -422,7 +440,137 @@ func recoverBPFState(eBPFSDKClient goelf.BpfSDKClient, policyEndpointeBPFContext
 		}
 	}
 
-	return isConntrackMapPresent, isPolicyEventsMapPresent, eventsMapFD, nil
+	//If update required, cleanup probes and gather data to re attach prpbes with new programs
+	if updateIngressProbe || updateEgressProbe {
+		// Get all loaded programs and maps
+		bpfState, err := eBPFSDKClient.GetAllBpfProgramsAndMaps()
+		if err != nil {
+			log.Info("GetAllBpfProgramsAndMaps failed: ", "error: ", err)
```
M00nF1sh, line 448: shall we return a error here instead of continue?

pkg/ebpf/bpf_client.go
```diff
@@ -422,7 +440,137 @@ func recoverBPFState(eBPFSDKClient goelf.BpfSDKClient, policyEndpointeBPFContext
...
+
+		// Get attached progIds
+		interfaceToIngressProgIds, interfaceToEgressProgIds, err := bpfTCClient.GetAllAttachedProgIds()
+		log.Info("Got attached ", "ingressprogIds ", len(interfaceToIngressProgIds), " egressprogIds ", len(interfaceToEgressProgIds))
+
+		//cleanup all existing filters
+		cleanupErr := bpfTCClient.CleanupQdiscs(updateIngressProbe, updateEgressProbe)
+		if cleanupErr != nil {
+			log.Info("Probe cleanup failed ", "error: ", cleanupErr)
+			sdkAPIErr.WithLabelValues("CleanupQdiscs").Inc()
```
M00nF1sh, line 468: again, shall we return error if failed to cleanup? what kind of condition will cause it fail and will it impact following up actions?

pkg/ebpf/bpf_client.go
```diff
@@ -422,7 +440,137 @@ func recoverBPFState(eBPFSDKClient goelf.BpfSDKClient, policyEndpointeBPFContext
...
+	}
+
+	return isConntrackMapPresent, isPolicyEventsMapPresent, eventsMapFD, interfaceNametoIngressPinPath, interfaceNametoEgressPinPath, nil
+}
+
+func (l *bpfClient) ReAttachEbpfProbes() error {
+	var networkPolicyMode string
+	var err error
+
+	// If we have any links for which we need to reattach the probes, fetcg NP mode from ipamd
```
M00nF1sh, line 493: nit: fetch

pkg/ebpf/bpf_client.go
```diff
@@ -422,7 +440,137 @@ func recoverBPFState(eBPFSDKClient goelf.BpfSDKClient, policyEndpointeBPFContext
...
+func (l *bpfClient) GetNetworkPolicyModeFromIpamd() (string, error) {
+
+	ctx := context.Background()
+	grpcLogger := ctrl.Log.WithName("grpcLogger")
+
+	// grpc connection waits till the ipmad is up and running
+	grpcLogger.Info("Trying to establish GRPC connection to ipamd")
+	grpcConn, err := rpcclient.New().Dial(ctx, LOCAL_IPAMD_ADDRESS, rpcclient.GetDefaultServiceRetryConfig(), rpcclient.GetInsecureConnectionType())
+	if err != nil {
+		grpcLogger.Error(err, "Failed to connect to ipamd")
```
M00nF1sh, line 562: we are not handling the error here, shouldn't we return? since grpcConn cannot be used

pkg/ebpf/bpf_client.go
```diff
@@ -422,7 +440,137 @@ func recoverBPFState(eBPFSDKClient goelf.BpfSDKClient, policyEndpointeBPFContext
 		}
 	}
 
-	return isConntrackMapPresent, isPolicyEventsMapPresent, eventsMapFD, nil
+	//If update required, cleanup probes and gather data to re attach prpbes with new programs
```
jayanthvn, line 443: typo - 'probes'

pkg/ebpf/bpf_client.go
```diff
@@ -422,7 +440,137 @@ func recoverBPFState(eBPFSDKClient goelf.BpfSDKClient, policyEndpointeBPFContext
...
+
+		if utils.IsStrictMode(networkPolicyMode) {
+			l.logger.Info("Updating pod_state map to default_deny for ", "podIdentifier: ", podIdentifier)
+			err = l.UpdatePodStateEbpfMaps(podIdentifier, DEFAULT_DENY, true, false)
+			if err != nil {
+				l.logger.Info("Map update(s) failed for, ", "podIdentifier ", podIdentifier, "error: ", err)
+			}
+		} else {
+			l.logger.Info("Updating pod_state map to default_allow for ", "podIdentifier: ", podIdentifier)
+			err = l.UpdatePodStateEbpfMaps(podIdentifier, DEFAULT_ALLOW, true, false)
```
jayanthvn, line 520: So we just allow reconcile request to fix it in future?

pkg/ebpf/bpf_client.go
```diff
@@ -422,7 +440,137 @@ func recoverBPFState(eBPFSDKClient goelf.BpfSDKClient, policyEndpointeBPFContext
...
+		}
+	}
+
+	for interfaceName, pinPath := range l.interfaceNametoEgressPinPath {
+		podIdentifier, _ := utils.GetPodIdentifierFromBPFPinPath(pinPath)
+		l.logger.Info("ReattachEbpfProbes ", "attaching egress for ", podIdentifier, "interface ", interfaceName)
+		_, err := l.attachEgressBPFProbe(interfaceName, podIdentifier)
+		if err != nil {
+			l.logger.Info("Failed to Attach Egress TC probe for", "interface: ", interfaceName, " podidentifier", podIdentifier)
+			sdkAPIErr.WithLabelValues("attachEgressBPFProbe").Inc()
```
jayanthvn, line 533: If probes fail to recover, should we consider returning init failed?

pkg/ebpf/bpf_client.go
Pavani-Panakanti: updated

pkg/ebpf/bpf_client.go
Pavani-Panakanti: updated

pkg/ebpf/bpf_client.go
Pavani-Panakanti: updated

pkg/ebpf/bpf_client.go
Pavani-Panakanti: Thanks for catching this. Fixed

pkg/ebpf/bpf_client.go
Pavani-Panakanti: We reconcile almost immediately on the restart as we are not waiting for any new pod events. it might be better to allow it to be fixed with reconcile request rather than failing on init and letting it go in crash loop

pkg/ebpf/bpf_client.go
Pavani-Panakanti: Yeah the reconcile request is handled almost immediately on the restart, so we can let reconcile request handle if incase there are any errors here

jayanthvn:
@Pavani-Panakanti - Lets fix the tests, before merging

jayanthvn (approved):

PR #367, opened 2025-02-12
Find the host veth name before attaching or detaching probes

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
This change finds the host veth pair of the pod before trying to attach or detach the traffic probes. Since the pod created by vpc-cni can have `vlan` (SGP in strict mode) or `eni` (SGP in standard or a regular pod), we loop through both the prefixes to check if the interface with that name exists. If either of them are not found, we return empty string 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/golang/mock v1.6.0
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.6.0
+	github.com/hashicorp/go-multierror v1.1.1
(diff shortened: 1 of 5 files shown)


jayanthvn:
Looks like build is failing

jaydeokar:
Depends on PR #365 since there have been changes that are added to the ebpf sdk.

jayanthvn (approved):
lgtm

Pavani-Panakanti (approved):

orsenthil (commented):
LGTM.

PR #376, opened 2025-03-12
Fix ebpf maps/progs cleanup test 

Pavani-Panakanti:
*Issue #, if available:*
With recent changes in NP, every pod will have maps/progs created even if policies are not applied to it. Once we run cyclonus tests and clean up the pods and policies, we run a pod which runs cleanup script to check if there are any leaked maps or progs on the node. As we are creating a new pod for this, there will be maps/progs created for this pod.

*Description of changes:*
 Exclude maps/progs with cleanup pod name while checking for leaked maps/progs. core dns pods will also have default probes attached as it is a deployment. Excluding those too 


jaydeokar (approved):
Maybe we can keep this as an input arg in future and keep the logic very basic. 
Looks good to me for now

PR #400, opened 2025-05-05
refactor policyendpoints controller

viveksb007:
*Issue #, if available:*

*Description of changes:* Refactor policy endpoint controller to be able to write UT for Reconcile method. In follow up PRs, will introduce a mock for `bpfclient`


Pavani-Panakanti:
@viveksb007 Can you add more details to the PR description. Thanks
Changes made
Why we need these changes 
testing details 
I see we already have a mock bpfclient. Do you meant adding more mock functions to that client ?

viveksb007:
@Pavani-Panakanti sure, I will run integ test and add details.

Main idea is to be able to test `Reconcile` method of `policyendpoint_reconciler.go`. I saw the Mock fn and don't think its a mock. This is current `NewMockBpfClient` (https://github.com/aws/aws-network-policy-agent/blob/d85dbabad07b2919422c0f95172854da75e3c087/pkg/ebpf/bpf_client_mock.go#L11) which return a real ebpfClient which is instantiated with dummy values but we can't actual mock any method of it. If we call `AttacheBPFProbes` fn of it, the test will result in `panic`.

With this refactor change, I have uplifted the client creation in `main.go` and simplified the Reconciler constructor. We can now create a `bpfMockClient` which will implement actual `BpfClient` interface methods for UT purpose. Then we can call Reconcile method of PolicyEndpoint controller in UT and write comprehensive UTs.

PR #403, opened 2025-05-07
minor script fixes

viveksb007:
*Issue #, if available:*

*Description of changes:*
Minor script fixes and improvements

- fixed readme format
- fixed `eksctl` call with region (without region call fails with following error)
```
 ~  eksctl utils describe-cluster-versions                                                                                                                                            ✔  viveksbh@dev-dsk-viveksbh-1d-59b3746d  22:01:31
Error: AWS Region must be set, please set the AWS Region in AWS config file or as environment variable
```
- added validation for installed binaries which are used when test is run

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -3,6 +3,7 @@ set -Eeuox pipefail
 
 DIR=$(cd "$(dirname "$0")"; pwd)
 
+source ${DIR}/lib/common.sh
 source ${DIR}/lib/cleanup.sh
 source ${DIR}/lib/cloudwatch.sh
 source ${DIR}/lib/cluster.sh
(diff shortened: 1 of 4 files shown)


scripts/run-tests.sh
```diff
@@ -26,6 +27,11 @@ cleanup() {
 
 trap cleanup EXIT
 
+check_is_installed docker
```
jaydeokar, line 30: docker should not be required ?

yash97:
Can we please make commit and PR description more descriptive on what enhancement and fix this PR is proposing. TIA!

jaydeokar (approved):

scripts/run-tests.sh
viveksb007: yeah, that is not needed.

we probably should have some enhancements where we can also `build_and_push` code changed images and run test on those (similar to vpc-cni repo)

PR #405, opened 2025-05-08
Fix Ginkgo command flag placement in run-cyclonus-tests.sh

yash97:
*Issue #, if available:*

*Description of changes:*

The Ginkgo command was failing with "Malformed arguments - detected a flag after the package list" error because the --no-color and --fail-on-pending flags were placed after the test binary path. This commit fixes the issue by moving these flags before the test binary path, following Ginkgo's required command structure where all flags must appear before the package list.

Looks like this error was introduced in 2.23.1 https://github.com/onsi/ginkgo/releases/tag/v2.23.1


jaydeokar (approved):
lgtm

PR #410, opened 2025-05-19
fix - avoid bpfcontext mangling

viveksb007:
*Issue #, if available:*

currently there is risk of assigning egressPgm of 1 `podId` to another `podId`. This will happen in during recovery cases where only ingressPgm is present for a `PodId`.

*Description of changes:*
This change moves the initialization in scope of each `bpfState`, such that we don't mangle bpfState from 1 PodId to another.

### DIFF
--- a/pkg/ebpf/bpf_client_test.go
+++ b/pkg/ebpf/bpf_client_test.go
@@ -9,6 +9,7 @@ import (
 	goelf "github.com/aws/aws-ebpf-sdk-go/pkg/elfparser"
 	goebpfmaps "github.com/aws/aws-ebpf-sdk-go/pkg/maps"
 	goebpfprogs "github.com/aws/aws-ebpf-sdk-go/pkg/progs"
+	"github.com/samber/lo"
 
 	mock_bpfclient "github.com/aws/aws-ebpf-sdk-go/pkg/elfparser/mocks"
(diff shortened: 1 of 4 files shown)


pkg/ebpf/bpf_client_test.go
```diff
@@ -706,58 +707,122 @@ func TestRecoverBPFState(t *testing.T) {
...
+			want: want{
+				isPolicyEventsMapPresent: true,
+				isConntrackMapPresent:    false,
+				eventsMapFD:              3,
+				bpfContextCount:          1,
+			},
+			wantErr: nil,
+		},
+		{
+			name:               "Prevent BpfContext mangling",
```
Pavani-Panakanti, line 795: Test name here might be confusing without context of this PR or issue. Can we rename it to something generic like " Recovering bpf programs "

Pavani-Panakanti (approved):

pkg/ebpf/bpf_client_test.go
viveksb007: I intended to name it like this as this test is validating conditions against the bug of assigning bpf prog of 1 PodId to another. 

I can leave it like this or rename it as `BpfContext Map validation post Bpf programs recovery` but something generic doesn't tell the reader what exactly are we testing here, as for all test cases we are calling `recoverBPFState()` and doing some validations.

Issue #417, opened 2025-06-09
High CPU usage during pod churns when NP targets large rules


viveksb007:
**What happened**:
NPA CPU utilization is high during high pod churning load in system.

**What you expected to happen**:
NPA should handle the incremental updates in NP efficiently.

**How to reproduce it (as minimally and precisely as possible)**:

> Test scale:
Tested with commit https://github.com/aws/aws-network-policy-agent/commit/92a9c0f9048b30a4f118df677ba99e0f15e71dc2 and following flags to prevent CPU spikes due to excessive logging
```
- args:
        - --enable-ipv6=false
        - --enable-network-policy=true
        - --enable-cloudwatch-logs=false
        - --enable-policy-event-logs=false
        - --log-level=error
        - --log-file=/var/log/aws-routed-eni/network-policy-agent.log
        - --metrics-bind-addr=:8162
        - --health-probe-bind-addr=:8163
        - --conntrack-cache-cleanup-period=300
        env:
        - name: AWS_EBPF_SDK_LOGLEVEL
          value: error
```

- Cluster with 100 nodes (`m5.large` instance type).
- Created 2 deployments in separate ns (`ns1` and `ns2`) which will each deploy 1000 pods.
- Created 4 similar NP per ns where each NP creates ingress/egress rules targeting other ns
```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: test-np
spec:
  podSelector: {}
  egress:
  - to:
    - namespaceSelector:
        matchLabels:
          name: ns1
  - to:
    - namespaceSelector:
        matchLabels:
          name: ns2
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          name: ns1
  - from:
    - namespaceSelector:
        matchLabels:
          name: ns2
  policyTypes:
  - Egress
  - Ingress
```
Created a pod churn of 300 pods every 30sec. 

Created 4 more similar NP per ns, so total of 8 NP per ns. Can see the CPU spiked from before.

### CPU Utilization:

<img width="1084" alt="Image" src="https://github.com/user-attachments/assets/621bf601-2b1a-4ceb-a1be-af79d7247ae6" />

We have profiled and identified some optimization areas in the reconciliation logic. We are working to add those optimizations and bring the CPU usage down during Pod churn load.

viveksb007:
Flame graph related to profiling the code on above scale.

[...]

PR #423, opened 2025-06-16
using in-mem ebpf map for syscall reduction

viveksb007:
*Issue #, if available:*
High CPU usage https://github.com/aws/aws-network-policy-agent/issues/417

This PR solved part of the problem to optimize ebpf map writing

*Description of changes:*
- upgraded ebpf-go dependency for fd re-use
- in-mem ebpf to make incremental changes to ebpf map

### DIFF
--- a/pkg/ebpf/bpf_client.go
+++ b/pkg/ebpf/bpf_client.go
@@ -143,6 +143,8 @@ func NewBpfClient(nodeIP string, enablePolicyEventLogs, enableCloudWatchLogs boo
 		EgressProgToPodsMap:       new(sync.Map),
 		AttachProbesToPodLock:     new(sync.Map),
 		DeletePodIdentifierLock:   new(sync.Map),
+		ingressInMemoryMap:        make(map[string]*InMemoryBpfMap),
+		egressInMemoryMap:         make(map[string]*InMemoryBpfMap),
 	}
 	ingressBinary, egressBinary, eventsBinary,
 		cliBinary, hostMask := TC_INGRESS_BINARY, TC_EGRESS_BINARY, EVENTS_BINARY, EKS_CLI_BINARY, IPv4_HOST_MASK
@@ -189,7 +191,7 @@ func NewBpfClient(nodeIP string, enablePolicyEventLogs, enableCloudWatchLogs boo
 	var interfaceNametoIngressPinPath map[string]string
 	var interfaceNametoEgressPinPath map[string]string
 	eventBufferFD := 0
-	isConntrackMapPresent, isPolicyEventsMapPresent, eventBufferFD, interfaceNametoIngressPinPath, interfaceNametoEgressPinPath, err = recoverBPFState(ebpfClient.bpfTCClient, ebpfClient.bpfSDKClient, ebpfClient.policyEndpointeBPFContext,
+	isConntrackMapPresent, isPolicyEventsMapPresent, eventBufferFD, interfaceNametoIngressPinPath, interfaceNametoEgressPinPath, err = ebpfClient.recoverBPFState(ebpfClient.bpfTCClient, ebpfClient.bpfSDKClient, ebpfClient.policyEndpointeBPFContext,
 		ebpfClient.GlobalMaps, ingressUpdateRequired, egressUpdateRequired, eventsUpdateRequired)
 	if err != nil {
 		//Log the error and move on
@@ -315,6 +317,10 @@ type bpfClient struct {
 	interfaceNametoEgressPinPath map[string]string
 	// Stores podIdentifier to deletepod lock mapping
 	DeletePodIdentifierLock *sync.Map
+	// This is in-memory map backed by ingressBpfMap
+	ingressInMemoryMap map[string]*InMemoryBpfMap
+	// This is in-memory map backed by egressBpfMap
+	egressInMemoryMap map[string]*InMemoryBpfMap
 	// FirewallRuleProcessor is used to convert firewall rules into ebpf Map content
 	fwRuleProcessor *fwrp.FirewallRuleProcessor
 }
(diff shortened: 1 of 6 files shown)


pkg/ebpf/bpf_client.go
```diff
@@ -803,25 +830,47 @@ func (l *bpfClient) UpdateEbpfMaps(podIdentifier string, ingressFirewallRules []
...
+			ingressProgFD := ingressProgInfo.Program.ProgFD
+			mapToUpdate := ingressProgInfo.Maps[TC_INGRESS_MAP]
 			log().Infof("Pod has an Ingress hook attached. Update the corresponding map progFD: %d, mapName: %s", ingressProgFD, TC_INGRESS_MAP)
-			err := l.updateEbpfMap(mapToUpdate, ingressFirewallRules)
-			duration := msSince(start)
-			sdkAPILatency.WithLabelValues("updateEbpfMap-ingress", fmt.Sprint(err != nil)).Observe(duration)
+
+			log().Infof("ingressInMem map %+v, bpfMap %+v", l.ingressInMemoryMap, mapToUpdate)
+			if _, exists := l.ingressInMemoryMap[podIdentifier]; !exists {
+				log().Infof("didn't found ingress in-mem map for %v", podIdentifier)
```
Pavani-Panakanti, line 839: Reword the log here for readability 
"Ingress in-mem map not found for "

pkg/ebpf/bpf_client.go
```diff
@@ -803,25 +830,47 @@ func (l *bpfClient) UpdateEbpfMaps(podIdentifier string, ingressFirewallRules []
...
+				l.ingressInMemoryMap[podIdentifier], err = NewInMemoryBpfMap(&mapToUpdate)
+				if err != nil {
+					log().Errorf("got err for ingress in-mem map for %v, err %v", podIdentifier, err)
+				} else {
+					log().Infof("ingress map for %v loaded successfully", podIdentifier)
+				}
+			}
+
+			err := l.updateEbpfMap(ingressFirewallRules, l.ingressInMemoryMap[podIdentifier])
```
Pavani-Panakanti, line 848: what will be the behavior if we error out above at line 841 and still procees with updateebpfmap here ?

pkg/ebpf/bpf_mem_map.go
```diff
@@ -0,0 +1,275 @@
...
+
+	log().Infof("Starting loadFromKernel operation")
+
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	// Clear current contents
+	m.contents = make(map[string][]byte)
+
+	// todo - ensure ebpf-sdk-go dependency is updated for correct behavior
```
Pavani-Panakanti, line 58: This is about the new PR to reuse fd ?

pkg/ebpf/bpf_mem_map.go
```diff
@@ -0,0 +1,275 @@
...
+		m.contents[key] = value
+	}
+
+	log().Infof("Loaded %d entries from kernel map", len(m.contents))
+	return nil
+}
+
+// Get retrieves a value from the in-memory map
+// Note: The returned byte slice should not be modified by the caller
+func (m *InMemoryBpfMap) Get(key string) ([]byte, bool) {
```
Pavani-Panakanti, line 89: I see we are not using lot of functions in this file. Its confusing and I think we should just keep the ones we need and add new func later if required ?

pkg/ebpf/bpf_mem_map.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	// Delete from in-memory representation
+	delete(m.contents, key)
+
+	return nil
+}
+
+// BulkUpdate efficiently updates multiple entries with optimized kernel updates
+func (m *InMemoryBpfMap) BulkUpdate(updates map[string][]byte) error {
```
Pavani-Panakanti, line 139: When will we need this ? We might always just use BulkRefresh right ?

pkg/ebpf/bpf_mem_map.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Apply updates to kernel
+	for k, v := range toAdd {
+		keyByte := []byte(k)
+		keyPtr := uintptr(unsafe.Pointer(&keyByte[0]))
+		valuePtr := uintptr(unsafe.Pointer(&v[0]))
+
+		if err := m.bpfMap.UpdateMapEntry(keyPtr, valuePtr); err != nil {
+			log().Errorf("Failed to update kernel map during bulk refresh for key %s: %v", k, err)
+			return err
```
Pavani-Panakanti, line 201: To confirm - even when we error out in between the in mem map and kernel map will be in sync ? and on next update we will try to diff again and fill that

pkg/ebpf/bpf_client.go
```diff
@@ -803,25 +830,47 @@ func (l *bpfClient) UpdateEbpfMaps(podIdentifier string, ingressFirewallRules []
...
+			mapToUpdate := ingressProgInfo.Maps[TC_INGRESS_MAP]
 			log().Infof("Pod has an Ingress hook attached. Update the corresponding map progFD: %d, mapName: %s", ingressProgFD, TC_INGRESS_MAP)
-			err := l.updateEbpfMap(mapToUpdate, ingressFirewallRules)
-			duration := msSince(start)
-			sdkAPILatency.WithLabelValues("updateEbpfMap-ingress", fmt.Sprint(err != nil)).Observe(duration)
+
+			log().Infof("ingressInMem map %+v, bpfMap %+v", l.ingressInMemoryMap, mapToUpdate)
+			if _, exists := l.ingressInMemoryMap[podIdentifier]; !exists {
+				log().Infof("didn't found ingress in-mem map for %v", podIdentifier)
+				l.ingressInMemoryMap[podIdentifier], err = NewInMemoryBpfMap(&mapToUpdate)
```
Pavani-Panakanti, line 840: Can we also evaluate once what happens when we reach here from two concurrent go routines

Pavani-Panakanti:
Looks good to me. Can you review the newly added logs ones and if we can move any of them to debug lets do it. Want to make sure we don't increase logging again eventually with new changes

viveksb007:
> Looks good to me. Can you review the newly added logs ones and if we can move any of them to debug lets do it. Want to make sure we don't increase logging again eventually with new changes

newly added logs are all called during in-mem map `init` cases so shouldn't be noisy. Only `BulkRefresh **` is called whenever Ebpf map is updated.

PR #434, opened 2025-07-03
moving noisy logs to debug

viveksb007:
*Issue #, if available:*
these logs are part of business logic which is executed frequently and results in high log volume if load (more NP/rules per NP)

*Description of changes:*
moving noisy logs to debug


Pavani-Panakanti:
These logs are imp for debugging to make sure IP and ports are updated in bpf map. Moving them to debug we might lose that info

PR #454, opened 2025-08-11
Log denys at highest log level

Pavani-Panakanti:
*Issue #, if available:*
DENY's should always be logged irrespective of log level. Move them to highest log level

*Description of changes:*

### DIFF
--- a/pkg/ebpf/events/events.go
+++ b/pkg/ebpf/events/events.go
@@ -218,7 +218,7 @@ func capturePolicyEvents(ctx context.Context, ringbufferdata <-chan []byte, enab
 				if rb.Verdict == VerdictDeny {
 					dropCountTotal.WithLabelValues(direction).Add(float64(1))
 					dropBytesTotal.WithLabelValues(direction).Add(float64(rb.PacketSz))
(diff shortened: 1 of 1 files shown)


pkg/ebpf/events/events.go
```diff
@@ -218,7 +218,7 @@ func capturePolicyEvents(ctx context.Context, ringbufferdata <-chan []byte, enab
 				if rb.Verdict == VerdictDeny {
 					dropCountTotal.WithLabelValues(direction).Add(float64(1))
 					dropBytesTotal.WithLabelValues(direction).Add(float64(rb.PacketSz))
-					log().Infof("Flow Info: Src IP: %s Src Port: %d Dest IP: %s Dest Port: %d Proto: %s Verdict: %s Direction: %s", utils.ConvByteToIPv6(rb.SourceIP).String(), rb.SourcePort,
+					log().Errorf("Flow Info: Src IP: %s Src Port: %d Dest IP: %s Dest Port: %d Proto: %s Verdict: %s Direction: %s", utils.ConvByteToIPv6(rb.SourceIP).String(), rb.SourcePort,
```
yash97, line 221: but it is not error right? This will provide incorrect signal during troubleshooting.

yash97: Ideally we should update this interface to support `V` method where we can optionally set log level.

viveksb007: agreed. if we want to log DENY irrespective of log level, we should de-couple it from current log config which is being set from outside (would a new logger whose config is explicitly set in code work?)

PR #458, opened 2025-08-27
Fix logger error in container logs

Pavani-Panakanti:
*Issue #, if available:*
All application logging has stack depth of 2 for caller. For the controller logger we have stack depth of 0. As we have set global caller stack depth of 2 so it is logging an error in container logs for creating logger for controller runtime
```
2025-08-27 04:20:32.080917205 +0000 UTC Logger.check error: failed to get caller
```

*Description of changes:*
For controller logger use stack depth of 0 for caller. Rest all application logging uses stack depth of 2

*Testing*
Above error does not appear in container log after the fix 

### DIFF
--- a/controllers/policyendpoints_controller.go
+++ b/controllers/policyendpoints_controller.go
@@ -39,10 +39,6 @@ import (
 	networking "k8s.io/api/networking/v1"
 )
 
-const (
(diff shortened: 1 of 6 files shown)


jaydeokar (approved):
lgtm

viveksb007:
closing this as its fixed in https://github.com/aws/aws-network-policy-agent/pull/503

PR #468, opened 2025-09-24
policy setup latency duration should be per pod

yash97:
*Issue #, if available:*

*Description of changes:*


yash97:
does not have to be per pod. But right now is middle of 2 things. If you see we register observation for this metric in [for loop](https://github.com/aws/aws-network-policy-agent/pull/468/files#diff-69e696b183818dd536602b0f86b920bfe39546a5d264ff85b9f4e3656a9e4767R260). So this means we take our duration contains, deriving pod targets time, and then in loop setting up ebpf for the pod. now this duration for second pod in the loop will be this plus time it took for first pod to setup. So either we should keep observe outside of loop so we get policy setup time which will contain time duration to configure all pods affected by that policy or just keep it per pod where we loose time to derive all target pods processing.

viveksb007:
agreed current metric is ambiguous and doesn't provide good signal. 

Option A makes sense to me. 

today, this latency can be impacted by other NP(s) which target the pods which this current NP is targetting.

Issue #482, opened 2025-11-14
Store last denied 5-tuples as an eBPF map


m00lecule:
Hello,

I would like to introduce an additional eBPF map storing the last N dropped tuples per each pod. Proposed eBPF map could be printed using `./aws-eks-na-cli ebpf dump-maps  <MAP_ID>` to allow the administrator confirm if traffic has been lately dropped by NP agent.
 
I was thinking to use `BPF_MAP_TYPE_QUEUE` type of map, however feel free to suggest something else.

Introducing such eBPF map wont increase the latency to allowlisted traffic, as the map insertion will be only triggered only on denied traffic.

Please also note I can volunteer to implement that by myself.

**What would you like to be added**:
Quite recently we had an production incident, caused by some traffic being dropped. During incident response we couldn't confirm what was the packet source IP due to confusing tunneling and NAT between the source and destination.

**Why is this needed**:
The vpc-cni NP agent is missing a lightweight tool to confirm if some traffic has been dropped lately.

Theoretically we could enable networkPolicyEventLogs, however we were hesitant as the environment is processing a huge volume of network traffic. Secondly the networkPolicyEventLogs requires at least 1000m cpu request, which seems like an additional operational risk, definitely a resource consuming operation.

viveksb007:
trying to understand the ask here. `Store last denied packets as an eBPF map` does this mean, you want to store the 5-tuples of the packet being dropped? or actual payload of the packet?

Issue #464, opened 2025-09-08
Add flow log metrics (DROP, ACCEPT, maybe count of open flows?)


chlunde:
**What would you like to be added**:
Metric counters for drops, optionally more metrics.

Ideally this could be very cheap compared to network flow logs, as only a few counters need to be exported, and it could use a map instead of a ring buffer, and be polled at whatever frequency the client desires.

**Why is this needed**:
We need to know if a network policy causes drops in our network. Some way to attribute drops to a workload, like a namespace label or a network policy name label, would also be required.

chlunde:
@yash97 Thanks, sorry for not noticing the metrics, either it was because the counter was 0 at the point or we hadn't upgraded the CNI. I see it works find now!

What do you think about
* per-namespace metrics and/or 
* how do you think the efficiency would be by having the BPF side only handling drops and not accept?

This warning is fairly scary:
> Notice: Enabling this feature requires one CPU core per node.

jaydeokar:
We are working on this and will add updates once we have something finalized

Issue #463, opened 2025-09-05
Optimize network-policy-agent CPU utilization when enablePolicyEventLogs set to true


m00lecule:
**What would you like to be added**:

Please optimize the networkpolicy event logging process (`enablePolicyEventLogs=true`) as after enabling it the CPU utilization spikes up to 2 vCPU.

According to official documentation such utilization is expected, however I have never encountered a Kubernetes CNI observability feature with such high resource requirements.  

 ref: https://docs.aws.amazon.com/eks/latest/userguide/network-policies-troubleshooting.html

**Why is this needed**:
So high CPU utilization is bearable during short debugging sessions, however it is not sufficient during longer time periods. Those essential metrics (number and volume of allowed/denied packets) are not available when networkpolicy logs are disabled. Such metrics are essential during production issues investigation, to conclude the traffic was or wasn't dropped by aws-network-policy-agent.

viveksb007:
@m00lecule i see you referring to increase in CPU spikes after enabling flag `enablePolicyEventLogs=true`. Could you share some details related to workload that is deployed on cluster and is there any churn and anything else going on in cluster during these spikes? 

this would help us in replication and figuring out which part is causing the spike. thanks

m00lecule:
It would be fine to introduce a parameter to only log denies from eBPF functions. As the main take away to be informed if NP agent has actually dropped the traffic without heavy cpu utilization.

viveksb007:
@m00lecule in this issue, you mentioned upto 2vCPU, do you what is typical CPU consumed by agent in your setup? if you can share some latest data related to it that will be helpful.

As there would be additional CPU consumed when this flag is enabled. I am trying to gauge that extra usage that this policy event logging is taking.

Issue #427, opened 2025-06-20
Filter out ACCEPT logs for cloudwatch logs


bdandoy:
**What would you like to be added**: 
Ability to filter logs sent to cloudwatch logs to only send DENY logs

**Why is this needed**:
Recently the logs on the server added the ability to not log ACCEPT event in https://github.com/aws/aws-network-policy-agent/pull/409. However we discovered this same filter was not applied to the cloudwatch logs.  This not only costs us in additional logging cost but it is also harder to see the logs we care about.

jaydeokar:
We recently added this with v1.3.7+ closing this

PR #491, opened 2025-11-30
feat: introduce policy event logs scope

m00lecule:
*Issues:*
https://github.com/aws/aws-network-policy-agent/issues/482
https://github.com/aws/aws-network-policy-agent/issues/463
https://github.com/aws/aws-network-policy-agent/issues/464#issuecomment-3291465539
https://github.com/aws/aws-network-policy-agent/issues/427
https://github.com/aws/aws-network-policy-agent/issues/601

*Description of changes:*
This feature allows to narrow the ebpf programs events scope (both `ACCEPT and DENY` or `only DENY`). To filter out the ACCEPT logs and metrics please specify  `--enable-policy-event-logs=true` and `--policy-event-logs-scope=deny`. 

The policy event logs volume can be overwhelming. The high CPU usage comes usually from userspace process processing the ebpf events published to `RINGBUF`, which are mostly the `ACCEPT` events.

Secondly the policy events logs are critical due to observability reasons. There is no command to confirm the NP is dropping traffic. The only way is to enable `enable-policy-event-logs=true`, which is quite resources heavy and might decrease environment performance.

Finally, by reducing the `ACCEPT` events we can get a guarantee that all `DENY` events are written to log files. The policy_events `RINGBUF` has a fixed size (https://github.com/aws/aws-network-policy-agent/blob/main/pkg/ebpf/c/v4events.bpf.c#L53), when exceeding the buffer capacity some old events are dropped and eventually missing in the log files.

Alternatively the feature could be implemented as:
- userspace logger logLevel, however it would still cause high CPU usage due to high volume processing caused by `ACCEPT` events,
- two RING_BUFFERS storing respectively `DENY` and `ACCEPT` events separately (probably cleaner approach, less ebpf maps lookups)
- configure the policy event logs scope per traffic type (egress, ingress) or even per replica set (configuration based on vpc-cni NP config and deployment annotations)

The goal is to discuss the overall architecture. Please consider this PR as prototype, not yet a final implementation.


m00lecule:
@viveksb007  kindly asking for an initial review

m00lecule:
@jayanthvn May I ask You to find someone to review this PR?

pkg/ebpf/bpf_client.go
```diff
@@ -230,6 +238,10 @@ func NewBpfClient(ctx context.Context, nodeIP string, enablePolicyEventLogs, ena
 			if mapName == AWS_EVENTS_MAP {
 				eventBufferFD = int(mapInfo.MapFD)
 			}
+			if mapName == AWS_EVENTS_SCOPE_MAP {
+				policyEventsScopeMap = mapInfo
+				isPolicyEventsScopeMapPresent = true
```
Pavani-Panakanti, line 243: why is this being set to true here ?

pkg/ebpf/bpf_client.go
```diff
@@ -248,6 +260,36 @@ func NewBpfClient(ctx context.Context, nodeIP string, enablePolicyEventLogs, ena
...
+		} else {
+			log().Errorf("Unable to get policyEventsScopeMap post recovery..error: %v", err)
+			sdkAPIErr.WithLabelValues("RecoveryFailed").Inc()
+			return nil, err
+		}
+
+		key := uint32(POLICY_EVENTS_SCOPE_MAP_KEY)
+		scope := uint8(utils.ACCEPT.Index())
+
+		if policyEventsLogsScope == "DENY" {
```
Pavani-Panakanti, line 278: there could be misconfigurations like "deny" we should validate the flag and log error if misconfigured, behavior can be fallback to default accept scope

pkg/ebpf/c/tc.v4egress.bpf.c
```diff
@@ -137,6 +141,15 @@ struct bpf_map_def_pvt SEC("maps") egress_pod_state_map = {
 
 struct bpf_map_def_pvt aws_conntrack_map;
 struct bpf_map_def_pvt policy_events;
+struct bpf_map_def_pvt policy_events_scope;
+
+static void publishPolicyEvent(struct data_t *evt) {	
+	__u32 plsc_key = 0;
+	struct policy_scope *plsc = bpf_map_lookup_elem(&policy_events_scope, &plsc_key);
```
Pavani-Panakanti, line 149: we are adding a new bpf map lookup in datapath. We should evaluate performance

Pavani-Panakanti:
Hey @m00lecule, Thanks for these changes. Overall problem and direction of gating events before putting then in ring buffer looks good. We need to add some unit tests and integration tests for these changes. As we are adding a bpf map and making bpf file changes, the changes needs some extensive testing with upgrades, downgrades and recovery with restarts. 
Can you also provide some scale numbers you are working with to see how much improvement this provides in cpu ?

m00lecule:
> Hey @m00lecule, Thanks for these changes. Overall problem and direction of gating events before putting then in ring buffer looks good. We need to add some unit tests and integration tests for these changes. As we are adding a bpf map and making bpf file changes, the changes needs some extensive testing with upgrades, downgrades and recovery with restarts. Can you also provide some scale numbers you are working with to see how much improvement this provides in cpu ?

cluster ~20 nodes and ~2000 pods. @Pavani-Panakanti will you takeover this PR (implement missing parts)? In general we don't want to see more than 300m cpu utilization when only DENY events are logged.

m00lecule:
@Pavani-Panakanti any updates?

pkg/ebpf/bpf_client.go
m00lecule: so the next if will pass

https://github.com/aws/aws-network-policy-agent/pull/491/changes/BASE..32dd6b7ce5b95a0ed2792086807c2a6e7b69dc99#diff-849e0e919d06071ad908c7294c3ccaa7b947701a796a59e5320c03191bb4c397R264

Pavani-Panakanti:
We are prioritizing this PR. We will have this in upcoming releases

pkg/ebpf/bpf_client.go
```diff
@@ -230,6 +238,10 @@ func NewBpfClient(ctx context.Context, nodeIP string, enablePolicyEventLogs, ena
 			if mapName == AWS_EVENTS_MAP {
 				eventBufferFD = int(mapInfo.MapFD)
 			}
+			if mapName == AWS_EVENTS_SCOPE_MAP {
+				policyEventsScopeMap = mapInfo
+				isPolicyEventsScopeMapPresent = true
+			}
```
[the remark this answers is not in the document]
Pavani-Panakanti: we should fix this. This will crash on fresh install and on binary upgrades during recovery path

Pavani-Panakanti:
@m00lecule Can you help address earlier comments and also copilot comments. 
All 3 CI checks are failing and unit tests not compiling - these needs to be addressed as well

m00lecule:
> @m00lecule Can you help address earlier comments and also copilot comments. All 3 CI checks are failing and unit tests not compiling - these needs to be addressed as well

Although i could implement those fixes i suggest you to take the ownership of this PR, given you might know more regards internal procedures.

Issue #489, opened 2025-11-25
"Logger.check error: failed to get caller" error in VPC CNI logs


kamil-clari:
**What happened**:
After the upgrade of AWS VPC CNI from version 1.19.5 to 1.20.4 in EKS cluster error log appeared in `aws-eks-nodeagent` container: `Logger.check error: failed to get caller`  
In VPC CNI 1.19.5 there is a 1.2.1 version of aws-network-policy-agent
IN VPC CNI 1.20.4 there is a 1.2.7 version of aws-network-policy-agent

EKS cluster in version 1.32.
Multiple similar issues was reported in comments: https://github.com/aws/aws-network-policy-agent/issues/247#issuecomment-3180525352

**Attach logs**
```
2025-11-25 14:27:04.511407263 +0000 UTC Logger.check error: failed to get caller
```

**What you expected to happen**:
Containers should start without any errors in logs

**How to reproduce it (as minimally and precisely as possible)**:
Run VPC CNI in version 1.20.0+ in EKS cluster and read logs from `aws-network-policy-agent`

**Environment**:
- Kubernetes version (use `kubectl version`): v1.32.9
- CNI Version 1.20.4
- Network Policy Agent Version 1.2.7

jaydeokar:
We need to review the caller and if we really need to use AddCallerSkip(2) in logging

kappa8219:
UPDATE again:

aws-network-policy-agent:v1.3.1 is still spamming in logs but it works!

viveksb007:
@kappa8219 those logs are red herring, they don't correspond to any real issue. I have raise the PR to fix that, logs on `stdout` should be fine after this PR is merged and released

jaydeokar:
Closing as the PR is merged

PR #495, opened 2025-12-10
fix: esolve failed to get caller error in GetControllerRuntimeLogger and add unit tests

HarshMehta112:
*Issue #489*

*Description of changes:*

**Root Cause**

Our structured logger configures zap with:

`zap.AddCallerSkip(2)`

This is required for handling wrapper logging methods (Info, Debug, etc.).
However, when creating a controller-runtime logger, we passed this zap logger directly into zapr.NewLogger().
Since zapr performs its own caller skip logic, the inherited +2 skip caused it to skip too many frames—resulting in zap failing to detect a valid caller and emitting repeated "failed to get caller" errors.

**✅ Fix**

We reset the caller skip offset before constructing the zapr logger:

```go 
zapLogger := zapSugared.Desugar().WithOptions(zap.AddCallerSkip(-2))
return zapr.NewLogger(zapLogger)
```

This removes the extra frames added for structured logging so that zapr can compute caller information correctly.

**Testing**

No "failed to get caller" messages appear after the fix.

Caller metadata is correctly populated for controller-runtime logs.

Unit tests confirm proper caller resolution:

`"caller":"logger/logger_test.go:21"`

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -53,8 +53,6 @@ func Get() Logger {
 
 func GetControllerRuntimeLogger() logr.Logger {
 	zapSugared := Get().(*structuredLogger).zapLogger
-	// Get the base zap logger without the AddCallerSkip that was added for structured logger methods
-	// zapr needs to calculate its own caller skip, so we start fresh
-	zapLogger := zapSugared.Desugar().WithOptions(zap.AddCallerSkip(-2))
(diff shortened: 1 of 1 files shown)


viveksb007:
@HarshMehta112 these change will effectively result in total CallerSkip of `3` as 2 are coming when `log = inputLogConfig.newZapLogger()` is called on line 41 and then we are adding 1 more on it.

I did something similar to the vpc-cni PR i linked above here https://github.com/aws/aws-network-policy-agent/pull/503

Issue #505, opened 2026-01-02
Support descibring which Kubernetes resource where the source IP or destination IP belongs to


andylim0221:
**What would you like to be added**:

The current source IP and destination IP is not direct enough to tell me which they are coming from in the logs. Ideally, we should have an idea on which Kubernetes workload and resource it's coming from, e.g. Deployment, ReplicaSet, Namespace etc.

**Why is this needed**:

Easier for troubleshooting. Example log format from GKE network policy logging:
https://docs.cloud.google.com/kubernetes-engine/docs/how-to/network-policy-logging#example_log_records

PR #506, opened 2026-01-10
chore: add metadata lookup hook to policy event logging

mohit7705:
This PR introduces a no-op metadata lookup hook in the policy event logging path
for both IPv4 and IPv6 flows.

There is no behavior change. This prepares the codebase for enriching policy logs
with Kubernetes workload metadata (Pod / Namespace / Workload), as discussed in #505.


viveksb007:
closing this PR as no update from author.

PR #531, opened 2026-03-23
ci: Build images natively per arch instead of under QEMU

yash97:
*Issue #, if available:* N/A

## Problem

PR CI builds the multi-arch image with QEMU emulation (`make docker-buildx`, both arches on one amd64 runner). Emulating arm64 on x86 makes the compiler run several times slower than native, and this one step dominates PR feedback time. The same QEMU path is used by every workflow that pushes images to ECR.

## What this PR does

**PR check build.** The docker build check now runs as a two-leg matrix: amd64 on `ubuntu-latest` and arm64 on `ubuntu-24.04-arm` (native Arm runners, free for public repos). No QEMU. A small aggregator job keeps the existing `Build Docker images` required-check name and fails, rather than skips, when either leg fails. GitHub treats a skipped required check as passing, so the `always()` guard matters.

**New composite actions.**
- `setup-ecr` logs in to ECR, makes sure the repository exists, and outputs the repository URI. Both actions below use it, so the login flow and the repository name live in one place.
- `build-and-push-image-per-arch` builds one architecture on a native runner and pushes it as `<tag>-<arch>`. It skips the build when the tag already exists, and an AWS API error aborts the step instead of being mistaken for "image exists".
- `create-multi-arch-manifest` assembles the `<tag>` index from the two per-arch images with `docker buildx imagetools create`.

**Workflows moved off QEMU.** Every image-pushing workflow now uses the new path:
- `e2e-conformance` runs both x86 (`t3.large`) and Graviton (`t4g.large`) nodes, so it builds both arches in parallel on native runners and then creates the multi-arch index in a separate job.
- `pr-cyclonus-tests` builds both arches in parallel the same way and pushes the index under the bare PR SHA, so the tested artifact keeps the same multi-arch coverage the QEMU path produced.
- `performance-tests` deploys only to x86 nodes (`c5.xlarge`), so it builds a single native amd64 image, tagged `<sha>-amd64`, and skips the arm64 build and the index. Nothing downstream assumes the bare `<sha>` tag.

The old `build-and-push-image` action is marked deprecated but kept for now, because an in-flight PR still references it. It gets deleted once that PR migrates.

**Guardrails.**
- `/test cyclonus` on a PR opened before this merges checks out a head that predates the new actions. The build job falls back to the default-branch copies in that case, so old PRs keep working without a rebase. The image is still built from the PR's source.
- The PR bot builds code that a maintainer approved but did not author, so its docker layer cache goes to separate `pr-amd64`/`pr-arm64` scopes. The trusted scheduled builds never import layers it cached.

## Measured build time (from this repo's Actions runs)

| Path | Wall time |
|---|---|
| QEMU `make docker-buildx` (both arches, one job) | 9m42s to 13m08s, median ~12m20s (6 recent runs) |
| Native per-arch matrix (parallel) | ~5m11s to 5m52s (runs on this branch) |

Same-commit A/B (run 23463150104): QEMU **15m40s** vs native amd64 **5m11s** / arm64 **4m31s** (~3x). Overall roughly 55-60% less wall time per image build, and every workflow that pushes images gets this, not just the PR check. Total runner minutes stay about the same.

## Testing

- The per-arch matrix jobs have run green on this branch.
- `actionlint` on all touched workflows (only pre-existing action-version warnings remain) and YAML validation on all four composite actions.
- The composite actions have not run end to end yet. Their first live exercise is the next `/test cyclonus` or the midnight conformance run after merge. If that is a concern, a manual `workflow_dispatch` of e2e-conformance right after merging verifies the full build-manifest-deploy path.


viveksb007:
@yash97 can you add PR description as to what this is supposed to do? is it reducing time for image build or something

PR #551, opened 2026-04-16
Removing CPE CRD checks as the CRD should be installed by default

jaydeokar:
*Issue #, if available:*
N/A

*Description of changes:*
This was introduced to protect the agent when CPE CRD is not installed in the cluster. However Network Policy Controller install this CRD by default via a reconciler, so this check should no longer be needed. 

Testing - 

TC1 - 
If NP is not enabled and the CRD is not installed, the controller just continues to run in background and doesn't crash
  

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,7 +22,6 @@ import (
 	"net/http"
 	_ "net/http/pprof"
 	"os"
-	"time"
 
 	cnirpc "github.com/aws/amazon-vpc-cni-k8s/rpc"
 	"github.com/aws/aws-network-policy-agent/pkg/ebpf"
@@ -39,7 +38,7 @@ import (
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
-	"k8s.io/client-go/discovery"
+
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
 	policyk8sawsv1 "github.com/aws/aws-network-policy-agent/api/v1alpha1"
@@ -104,33 +103,10 @@ func main() {
 
 	ctx := ctrl.SetupSignalHandler()
 
-	disc, err := discovery.NewDiscoveryClientForConfig(restCFG)
-	if err != nil {
-		log.Errorf("unable to create discovery client %v", err)
-		os.Exit(1)
-	}
 	var policyEndpointController *controllers.PolicyEndpointsReconciler
 	var clusterPolicyEndpointController *controllers.ClusterPolicyEndpointsReconciler
 	if ctrlConfig.EnableNetworkPolicy {
-		log.Info("Network Policy is enabled, checking CRDs...")
-		// Check if required CRDs exist. Need to remove this eventually when the CPE CRD is available everywhere
-		requiredGV := "networking.k8s.aws/v1alpha1"
-		requiredResource := "clusterpolicyendpoints"
-
-		skipCPEController := true
-
-		resources, err := disc.ServerResourcesForGroupVersion(requiredGV)
-		if err != nil {
-			log.Errorf("Cannot list resources in the cluster in GVK %s, err %v", requiredGV, err)
-		} else {
-			for _, r := range resources.APIResources {
-				if r.Name == requiredResource {
-					log.Infof("CRD found in the required GVK resource %s", requiredResource)
-					skipCPEController = false
-					break
-				}
-			}
-		}
+		log.Info("Network Policy is enabled, registering controllers...")
 
 		var nodeIP string
 		if !ctrlConfig.EnableIPv6 {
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -151,35 +127,14 @@ func main() {
...
-					}
-					log.Infof("CRD still not available, continuing to monitor: %s, %v", requiredResource, err)
-				}
-			}()
+
+		clusterPolicyEndpointController = controllers.NewClusterPolicyEndpointsReconciler(mgr.GetClient(), nodeIP, ebpfClient)
+		if err = clusterPolicyEndpointController.SetupWithManager(ctx, mgr); err != nil {
+			log.Errorf("unable to create controller ClusterPolicyEndpoints %v", err)
+			os.Exit(1)
 		}
```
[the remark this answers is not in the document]
jaydeokar: Controller runtime should give the error where it's unable to get an CRD for CPE. This is not required

pkg/rpc/rpc_handler_test.go
```diff
@@ -70,3 +72,93 @@ func TestEnforceNpToPod_ClearDeletedPodBeforeAttach(t *testing.T) {
...
+	resp, err := s.EnforceNpToPod(context.Background(), &rpc.EnforceNpRequest{
+		K8S_POD_NAME:        "nginx-abc123",
+		K8S_POD_NAMESPACE:   "default",
+		NETWORK_POLICY_MODE: "standard",
+		InterfaceCount:      1,
+	})
+	assert.NoError(t, err)
+	assert.True(t, resp.Success)
+	assert.Empty(t, mockBpfClient.CallLog)
+}
```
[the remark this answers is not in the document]
jaydeokar: EBPF client is shared, so if either CPE ebpf client or PE ebpf client is null, it should return success. It's a day 1 behavior

pkg/rpc/rpc_handler_test.go
```diff
@@ -70,3 +72,93 @@ func TestEnforceNpToPod_ClearDeletedPodBeforeAttach(t *testing.T) {
 	assert.Equal(t, "ClearDeletedPod", mockBpfClient.CallLog[1])
 	assert.Equal(t, "AttacheBPFProbes", mockBpfClient.CallLog[2])
 }
+
+func TestEnforceNpToPod_NilClusterPolicyReconciler_ReturnsSuccess(t *testing.T) {
+	mockBpfClient := &ebpf.MockBpfClient{}
+	reconciler := controllers.NewPolicyEndpointsReconciler(nil, "10.0.0.1", mockBpfClient, false)
+
+	s := &server{
```
viveksb007, line 80: this is invalid state of server instance correct?

post this change, we are making both either both reconciler or no reconciler. should we add if 1 reconciler is nil and other not nil in `RunRPCHandler` method and return err from there accordingly. And remove these tests which test single reconciler conditions which are not valid and might confuse readers. wdyt?

pkg/rpc/rpc_handler_test.go
```diff
@@ -70,3 +72,93 @@ func TestEnforceNpToPod_ClearDeletedPodBeforeAttach(t *testing.T) {
...
+	assert.True(t, resp.Success)
+	// Should early-return without calling any eBPF methods
+	assert.Empty(t, mockBpfClient.CallLog)
+}
+
+func TestEnforceNpToPod_NilClusterPolicyEBPFClient_ReturnsSuccess(t *testing.T) {
+	mockBpfClient := &ebpf.MockBpfClient{}
+	reconciler := controllers.NewPolicyEndpointsReconciler(nil, "10.0.0.1", mockBpfClient, false)
+	// clusterReconciler with nil eBPF client
+	clusterReconciler := controllers.NewClusterPolicyEndpointsReconciler(nil, "10.0.0.1", nil)
```
viveksb007, line 101: same for this test, is it testing a valid case? 

out ebpfClient will always be non-nil in main.go if clusterReconciler is being created 

`ebpfClient := lo.Must1`

viveksb007 (commented):
lgtm, few q on tests

pkg/rpc/rpc_handler.go
```diff
@@ -49,8 +49,9 @@ type server struct {
 
 // EnforceNpToPod processes CNI Enforce NP network request
 func (s *server) EnforceNpToPod(ctx context.Context, in *rpc.EnforceNpRequest) (*rpc.EnforceNpReply, error) {
-	// TODO: Add the clusterPolicyReconciler nil check here once cluster policies CRDs are available everywhere
-	if s.policyReconciler == nil || s.policyReconciler.GeteBPFClient() == nil {
+	// If NPA is disabled, controllers are nil. The gRPC server still runs to serve CNI requests,
+	// so return success without enforcing any policy.
+	if s.policyReconciler == nil || s.policyReconciler.GeteBPFClient() == nil || s.clusterPolicyReconciler == nil || s.clusterPolicyReconciler.GeteBPFClient() == nil {
```
yash97: we should make a fn which should return if any reconcilers are active or not.

pkg/rpc/rpc_handler_test.go
jaydeokar: Yeah makes sense to me. I will update this

pkg/rpc/rpc_handler.go
jaydeokar: Update this part for readability. LMK if you see any concerns

pkg/rpc/rpc_handler.go
```diff
@@ -49,17 +49,29 @@ type server struct {
...
 // EnforceNpToPod processes CNI Enforce NP network request
 func (s *server) EnforceNpToPod(ctx context.Context, in *rpc.EnforceNpRequest) (*rpc.EnforceNpReply, error) {
-	// TODO: Add the clusterPolicyReconciler nil check here once cluster policies CRDs are available everywhere
-	if s.policyReconciler == nil || s.policyReconciler.GeteBPFClient() == nil {
+	// If NPA is disabled, controllers are nil. The gRPC server still runs to serve CNI requests,
+	// so return success without enforcing any policy.
+	log().Infof("Received Enforce Network Policy Request for Pod: %s Namespace: %s Mode: %s", in.K8S_POD_NAME, in.K8S_POD_NAMESPACE, in.NETWORK_POLICY_MODE)
+	var err error
+	peReconcilerReady := s.policyReconciler != nil && s.policyReconciler.GeteBPFClient() != nil
+	cpeReconcilerReady := s.clusterPolicyReconciler != nil && s.clusterPolicyReconciler.GeteBPFClient() != nil
```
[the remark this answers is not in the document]
jaydeokar: Invalid case

pkg/rpc/rpc_handler.go
```diff
@@ -49,17 +49,29 @@ type server struct {
 
 // EnforceNpToPod processes CNI Enforce NP network request
 func (s *server) EnforceNpToPod(ctx context.Context, in *rpc.EnforceNpRequest) (*rpc.EnforceNpReply, error) {
-	// TODO: Add the clusterPolicyReconciler nil check here once cluster policies CRDs are available everywhere
-	if s.policyReconciler == nil || s.policyReconciler.GeteBPFClient() == nil {
+	// If NPA is disabled, controllers are nil. The gRPC server still runs to serve CNI requests,
+	// so return success without enforcing any policy.
+	log().Infof("Received Enforce Network Policy Request for Pod: %s Namespace: %s Mode: %s", in.K8S_POD_NAME, in.K8S_POD_NAMESPACE, in.NETWORK_POLICY_MODE)
```
Pavani-Panakanti: can we add this log only when np is enabled like earlier. Np will consume unnecessary cpu logging this for every pod when np is disabled

jaydeokar: Updated

Pavani-Panakanti (approved):

Issue #90, opened 2023-10-05
Allowing ICMP in network policy


ankenyr:
Currently it seems that ICMP is not allowed as a protocol in a network policy but it seems to be at least partially implemented and planned for.

https://github.com/aws/aws-network-policy-agent/blob/d764cafebca031cadd28eaa962e9af524edb2298/pkg/utils/utils.go#L21
https://github.com/aws/aws-network-policy-agent/blob/d764cafebca031cadd28eaa962e9af524edb2298/pkg/ebpf/events/events.go#L118

However when attempting to use protocol ICMP in the following way

```
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  namespace: cnitest
  name: demo
spec:
  podSelector:
    matchLabels: {}
  policyTypes:
  - Egress
  egress:
    - to:
        - ipBlock:
            cidr: 8.8.8.8/32
      ports:
        - protocol: ICMP
```

It results in an error message
`The NetworkPolicy "demo-app-deny-all" is invalid: spec.egress[0].ports[0].protocol: Unsupported value: "ICMP": supported values: "TCP", "UDP", "SCTP"`

This is an issue for any network using IPv6. [RFC4890](https://datatracker.ietf.org/doc/html/rfc4890#section-4.3.1) states that certain ICMP messages should always be allowed while others may be filtered. 

As soon as a filter is applied it negates all ICMP traffic unless the protocol is not specified and even then, only to the network or pods that are targeted. This makes it impossible to allow ICMP within the cluster while only allowing specific ports and protocols for direct communication.

achevuru:
ICMP isn't supported in the [Kubernetes Network policy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) resource as the above error indicates. For IPv6, we allow the ICMPv6 by default to allow Neighbor Discovery (NS/NA) packets.

ankenyr:
While it may not be listed in the K8s network policy that does not mean it cannot be supported. Is this something that AWS would expose to allow support for? It seems like most things are in place to allow this. My understanding is that most CNI's take the base K8s network policy as a starting point and then add to it.

achevuru:
[...]

PR #561, opened 2026-04-30
Allow IMCP fragmentation needed packets

### LINKED ISSUE #207: ICMP Type 3 messages being blocked on an already open connection.
**What happened**:
A call to an external SOAP service kept timing out; when using curl it returned a no response error message.
Once I worked out how to enable the flow logs (instructions missing from the online configuration guide) I saw this:
```
{"level":"info","ts":"2024-02-07T19:07:00.273Z","logger":"ebpf-client","msg":"Flow Info:  ","Src IP":"100.64.44.16","Src Port":48530,"Dest IP":"57.20.37.65","Dest Port":8443,"Proto":"TCP","Verdict":"ACCEPT"}
{"level":"info","ts":"2024-02-07T19:07:00.513Z","logger":"ebpf-client","msg":"Flow Info:  ","Src IP":"57.20.37.65","Src Port":0,"Dest IP":"100.64.44.16","Dest Port":0,"Proto":"ICMP","Verdict":"DENY"}
```
Once I saw the ICMP DENY message I was confused as this is not an Ingress connection; it is on the response to the already established Egress connection over which the SOAP request was sent.
When playing around with the NetworkPolicy definition for this app I found that the ICMP is accepted when I open an Ingress rule for the IP with no protocol or port restriction.

When using wireshark to take a network packet snapshot I saw the ICMP message is a Type 3 - Fragmentation needed.
I have been assured by our Networking team that such packets should be allowed over an open connection.

So, my issue is why is this ICMP Type 3 message being treated as a new Ingress when it is not; and why these messages are blocked when it seems they should be allowed?

**Attach logs**

I could not get that Log Collection tool to run on our Bottlerocket node.

The ICMP Type 3 message should be allowed to reach the pod as it is coming in on an already established Egress connection.

**How to reproduce it (as minimally and precisely as possible)**:
Make a call to a service that sends the following ICMP Type 3 message:
```
Internet Control Message Protocol
    Type: 3 (Destination unreachable)
    Code: 4 (Fragmentation needed)
    Checksum: 0xc2e1 [correct]
    [Checksum Status: Good]
    Unused: 0000
    MTU of next hop: 1372
```

**Anything else we need to know?**:
Previously we used Calico and with the exact same Network Policy definition and it works; using wireshark we can see the ICMP message reaching the pod. So Calico's implementation allows ICMP Type 3 messages on an already open Egress connection.

I also have an open AWS support case about this; they were the ones who suggested raising a bug report here.

**Environment**:
- Kubernetes version (use `kubectl version`):  v1.28.5-eks-5e0fdde
- CNI Version: v1.16.2-eksbuild.1
- Network Policy Agent Version: v1.0.7-eksbuild.1
- OS (e.g: `cat /etc/os-release`): "Amazon Linux 2"
- Kernel (e.g. `uname -a`): 6.1.61

wedge-jarrad:
*Issue #, if available:*

Fixes #207. This issue is already closed as a duplicate of #90 but I don't think it is the same thing. 207 is specifically about the failure of the path MTU discovery process due to all ICMP messages being blocked while 90 is asking for broad support for ICMP in network policies.

*Description of changes:*

Allow ICMP destination unreachable (fragmentation needed) packets which are required by the path MTU discovery process specified by [RFC1191](https://datatracker.ietf.org/doc/html/rfc1191).

### DIFF
--- a/pkg/ebpf/c/tc.v4ingress.bpf.c
+++ b/pkg/ebpf/c/tc.v4ingress.bpf.c
@@ -344,6 +344,7 @@ int handle_ingress(struct __sk_buff *skb)
 		struct tcphdr *l4_tcp_hdr = data + sizeof(struct iphdr);
 		struct udphdr *l4_udp_hdr = data + sizeof(struct iphdr);
 		struct sctphdr *l4_sctp_hdr = data + sizeof(struct iphdr);
+		struct icmphdr *icmp_hdr = data + sizeof(struct iphdr);
 
 		if (data + sizeof(*ip) > data_end) {
 			return BPF_OK;
@@ -373,6 +374,12 @@ int handle_ingress(struct __sk_buff *skb)
 				}
 				l4_src_port = (((((unsigned short)(l4_sctp_hdr->source) & 0xFF)) << 8) | (((unsigned short)(l4_sctp_hdr->source) & 0xFF00) >> 8));
 				l4_dst_port = (((((unsigned short)(l4_sctp_hdr->dest) & 0xFF)) << 8) | (((unsigned short)(l4_sctp_hdr->dest) & 0xFF00) >> 8));
+				break;
+			case IPPROTO_ICMP:
+				// Allow ICMP destination unreachable (fragmentation needed) packets
+				if (icmp_hdr->type == 3 && icmp_hdr->code == 4) {
+					return BPF_OK;
+				}
 				break;
 		}
 


viveksb007:
@wedge-jarrad thanks for contribution. But ICMP protocol type isn't supported in K8s Network Policy schema (https://kubernetes.io/docs/concepts/services-networking/network-policies/#network-traffic-filtering), so there is no way for the dataplane to allow ICMP traffic as the intent can't be specified using K8s Network Policy object itself.

Closing this PR. Let me know if there are additional questions

wedge-jarrad:
Hello @viveksb007 

No support for ICMP in the k8s network policy schema is required for this. The proposed change is to always accept a specific type of ICMP traffic to ensure that path MTU discovery works as described in RFC1191.

A similar approach was taken by the IPv6 implementation. It allows *all* ICMPv6 traffic to ensure proper network functionality.

https://github.com/aws/aws-network-policy-agent/blob/bda60c94aaca3eeb7393353d19594cdb02321f52/pkg/ebpf/c/tc.v6ingress.bpf.c#L355-L358

This PR only addresses the specific situation described in #207 where blocking ICMP type 3 code 4 packets prevents path MTU discovery from happening. Perhaps my mentioning #90 confused the issue. That issue is requesting full ICMP support in policy definitions. But that is not what I've set out to do here, nor is it necessary to solve this problem.

viveksb007:
if the NetworkPolicy is Deny all, this ICMP packet will still go through, right?

wedge-jarrad:
@viveksb007 I believe so, yes. The alternative (and what I had to do to work around the problems that this caused for me) is you have to allow **all** traffic from any destination which might send back an ICMP fragmentation needed packet. You essentially can't use network policies if this situation occurs in your environment.

Blindly accepting ICMP packets comes with some risk, but it seems that was considered acceptable for IPv6. This exception is much more targeted than what was done for IPv6.

The [RFC](https://datatracker.ietf.org/doc/html/rfc1191#section-8) itself describes some of the malicious activity that could result:

>    This Path MTU Discovery mechanism makes possible two denial-of-
>   service attacks, both based on a malicious party sending false
>   Datagram Too Big messages to an Internet host.
>
>   In the first attack, the false message indicates a PMTU much smaller
>  than reality.  This should not entirely stop data flow, since the
> victim host should never set its PMTU estimate below the absolute
>   minimum, but at 8 octets of IP data per datagram, progress could be
>   slow.
>
>   In the other attack, the false message indicates a PMTU greater than
>   reality.  If believed, this could cause temporary blockage as the
>   victim sends datagrams that will be dropped by some router.  Within
>   one round-trip time, the host would discover its mistake (receiving
>   Datagram Too Big messages from that router), but frequent repetition
>   of this attack could cause lots of datagrams to be dropped.  A host,
>   however, should never raise its estimate of the PMTU based on a
>   Datagram Too Big message, so should not be vulnerable to this attack.
>
>   A malicious party could also cause problems if it could stop a victim
>   from receiving legitimate Datagram Too Big messages, but in this case
>   there are simpler denial-of-service attacks available.

At present, AWS network policy is inadvertently acting as the "malicious party" described in the last paragraph.

PR #563, opened 2026-04-30
Fix attach-state collision in GetPodNamespacedName

parlakisik:
Pods ("ab","c") and ("a","bc") produced the same in-memory key "abc", causing isEBPFProbeAttached to false-positive and AttacheBPFProbes to skip TC hook installation for the second pod — a policy bypass.

"_" is forbidden in DNS-1123 names and namespaces, making the key injective. Change is in-process only; kernel state is keyed by GetPodIdentifier and unaffected, so no migration is required.

### DIFF
--- a/pkg/ebpf/bpf_client_test.go
+++ b/pkg/ebpf/bpf_client_test.go
@@ -100,6 +100,28 @@ func TestBpfClient_IsEBPFProbeAttached(t *testing.T) {
 	}
 }
 
+// Pre-attach pod ("ab", "c") and verify pod ("a", "bc") does not inherit
+// attachment state — the buggy raw concatenation aliased these to the same key.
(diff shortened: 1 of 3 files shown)


jaydeokar:
Overall looks good to me.. Can you confirm we testing upgrade/downgrade scenarios here ?

Pavani-Panakanti:
LGTM. Can you make sure to address above comment before merging. Thanks

Pavani-Panakanti (approved):

parlakisik:
The successful test flow is :
- creating pods with network policies  in previous NPA image .
- validate policies are working 
- update NPA image to the new one 
- validate NPA is running and policies are working and no error logs 
- revert back NPA image 
- validate NPA is running and policies are working and no error logs

PR #565, opened 2026-05-05
add liveness and readiness probe

yash97:
*Issue #, if available:*

*Description of changes:*

Testing, 
Tested readiness probe by removing grpc socket. Sample logs when socket is removed.

```
{"level":"error","ts":"2026-05-05T14:36:43.331Z","caller":"healthz/healthz.go:59","msg":"grpc-socket check failed: grpc health check failed: rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial unix /var/run/aws-node/npa.sock: connect: no such file or directory\""}
{"level":"info","ts":"2026-05-05T14:36:43.331Z","logger":"controller-runtime.healthz","caller":"healthz/healthz.go:128","msg":"healthz check failed","statuses":[{},{}]}
{"level":"error","ts":"2026-05-05T14:36:53.332Z","caller":"healthz/healthz.go:59","msg":"grpc-socket check failed: grpc health check failed: rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial unix /var/run/aws-node/npa.sock: connect: no such file or directory\""}
{"level":"info","ts":"2026-05-05T14:36:53.332Z","logger":"controller-runtime.healthz","caller":"healthz/healthz.go:128","msg":"healthz check failed","statuses":[{},{}]}
```

```
Add a gRPC Unix-socket health check that performs a gRPC Health/Check RPC against the agent.
Add readiness checks for bpffs correctness (mounted + writable) and presence of required pinned BPF maps.
Register the new checks with the controller manager and add unit tests for the eBPF readiness checks.
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -49,7 +49,6 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	//+kubebuilder:scaffold:imports
 )
 
@@ -140,18 +139,32 @@ func main() {
 
 	//+kubebuilder:scaffold:builder
 
-	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
-		log.Errorf("unable to set up health check %v", err)
+	// Liveness probe: gRPC server responsiveness.
+	if err := mgr.AddHealthzCheck("grpc-socket", rpc.NewGRPCSocketLivenessCheck(npaSocketPath)); err != nil {
+		log.Errorf("unable to set up gRPC socket health check %v", err)
 		os.Exit(1)
 	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
-		log.Errorf("unable to set up ready check %v", err)
+
+	// Readiness probe: gRPC server responsiveness and BPF map pins (when
+	// network policy is enabled).
+	if err := mgr.AddReadyzCheck("grpc-socket", rpc.NewGRPCSocketLivenessCheck(npaSocketPath)); err != nil {
+		log.Errorf("unable to set up grpc-socket readiness check %v", err)
 		os.Exit(1)
+	}
+
+	if ctrlConfig.EnableNetworkPolicy {
+		readyzPaths := []string{ebpf.CONNTRACK_MAP_PIN_PATH}
+		if ctrlConfig.EnablePolicyEventLogs {
+			readyzPaths = append(readyzPaths, ebpf.POLICY_EVENTS_MAP_PIN_PATH)
+		}
+		if err := mgr.AddReadyzCheck("bpf-maps", ebpf.NewGlobalMapsReadinessCheck(readyzPaths)); err != nil {
+			log.Errorf("unable to set up bpf-maps readiness check %v", err)
+			os.Exit(1)
+		}
 	}
 
 	// CNI makes rpc calls to NP agent regardless NP is enabled or not
 	// need to start rpc always
-	// todo: add a liveness probe to this gRPC server and remove closing based on this errCh, liveness probe will check and re-start this container
 	errCh, err := rpc.RunRPCHandler(policyEndpointController, clusterPolicyEndpointController, npaSocketPath)
 	if err != nil {
 		log.Errorf("Failed to set up gRPC Handler %v", err)
--- a/main.go
+++ b/main.go
@@ -134,8 +134,7 @@ func main() {
 			os.Exit(1)
 		}
 
-		readyzPaths := []string{ebpf.CONNTRACK_MAP_PIN_PATH}
-		readyzPaths = append(readyzPaths, ebpf.POLICY_EVENTS_MAP_PIN_PATH)
+		readyzPaths := []string{ebpf.CONNTRACK_MAP_PIN_PATH, ebpf.POLICY_EVENTS_MAP_PIN_PATH}
 		if err := mgr.AddReadyzCheck("bpf-maps", ebpf.NewGlobalMapsReadinessCheck(readyzPaths)); err != nil {
 			log.Errorf("unable to set up bpf-maps readiness check %v", err)
 			os.Exit(1)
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -139,18 +138,32 @@ func main() {
...
+		}
+		if err := mgr.AddReadyzCheck("bpf-maps", ebpf.NewGlobalMapsReadinessCheck(readyzPaths)); err != nil {
+			log.Errorf("unable to set up bpf-maps readiness check %v", err)
+			os.Exit(1)
+		}
+	}
+
 	// CNI makes rpc calls to NP agent regardless NP is enabled or not
 	// need to start rpc always
-	// todo: add a liveness probe to this gRPC server and remove closing based on this errCh, liveness probe will check and re-start this container
```
viveksb007, line 154: can we simplify the closing based on `errCh` happen currently? since we are adding liveness and readiness probes 

we don't need closing based on errCh now as liveness probe will re-start the container in case RPC handler is not running

main.go
```diff
@@ -139,18 +138,32 @@ func main() {
...
-		log.Errorf("unable to set up ready check %v", err)
+
+	// Readiness probe: gRPC server responsiveness and BPF map pins (when
+	// network policy is enabled).
+	if err := mgr.AddReadyzCheck("grpc-socket", rpc.NewGRPCSocketLivenessCheck(npaSocketPath)); err != nil {
+		log.Errorf("unable to set up grpc-socket readiness check %v", err)
 		os.Exit(1)
 	}
 
+	if ctrlConfig.EnableNetworkPolicy {
```
viveksb007: can we move this to above check `if ctrlConfig.EnableNetworkPolicy` where we are conditionally starting the controller?

main.go
yash97: right now keeping it as it is. Once We update the manifest of node agent with liveness probe i will remove it.

main.go
yash97: done

main.go
```diff
@@ -133,18 +132,35 @@ func main() {
 			log.Errorf("unable to create controller ClusterPolicyEndpoints %v", err)
 			os.Exit(1)
 		}
+
+		readyzPaths := []string{ebpf.CONNTRACK_MAP_PIN_PATH}
+		if ctrlConfig.EnablePolicyEventLogs {
```
viveksb007: can you double check, i think its supposed to be present irrespective of boolean `EnablePolicyEventLogs`

main.go
```diff
@@ -133,18 +132,35 @@ func main() {
...
 		log.Info("Network Policy is disabled, skip the controller registration")
 	}
 
 	//+kubebuilder:scaffold:builder
 
-	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
-		log.Errorf("unable to set up health check %v", err)
+	// The gRPC server runs whether or not network policy is enabled, so its
+	// health checks are registered unconditionally.
+	if err := mgr.AddHealthzCheck("grpc-socket", rpc.NewGRPCSocketHealthCheck(npaSocketPath)); err != nil {
```
viveksb007: for the agent enable case, above `AddReadyzCheck` will be called first, there will be 3 checkers registered against readyz, when /readyz is called, will all those be invoked and all their success will return true?

yash97: yes check this code of controller runtime which goes thorugh all checks https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/healthz/healthz.go#L45

main.go
```diff
@@ -134,18 +133,33 @@ func main() {
 			log.Errorf("unable to create controller ClusterPolicyEndpoints %v", err)
 			os.Exit(1)
 		}
+
+		readyzPaths := []string{ebpf.CONNTRACK_MAP_PIN_PATH}
```
viveksb007: can simplify as 

```
readyzPaths := []string{ebpf.CONNTRACK_MAP_PIN_PATH, ebpf.POLICY_EVENTS_MAP_PIN_PATH}
```

pkg/ebpf/healthcheck_test.go
```diff
@@ -0,0 +1,73 @@
...
+	t.Run("regular dir is not bpffs", func(t *testing.T) {
+		// A normal tmpfs/ext4 dir has a non-BPF_FS_MAGIC fs type, so the
+		// check must reject it. This is the realistic "bpffs failed to mount,
+		// agent is pinning into the root fs" failure mode.
+		err := NewBpfFsReadinessCheck(t.TempDir())(nil)
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "not a bpffs mount")
+	})
+
+	t.Run("real bpffs passes when mounted", func(t *testing.T) {
```
viveksb007: is this test giving any signal?

pkg/rpc/healthcheck.go
```diff
@@ -0,0 +1,72 @@
...
+func NewGRPCSocketHealthCheck(socketPath string) func(r *http.Request) error {
+	return func(r *http.Request) error {
+		baseCtx := context.Background()
+		if r != nil {
+			baseCtx = r.Context()
+		}
+		ctx, cancel := context.WithTimeout(baseCtx, livenessCheckTimeout)
+		defer cancel()
+
+		conn, err := grpc.NewClient(
```
viveksb007: can you double check if creating the client every time has any cost associated to it? otherwise we can create client once and call Check as part of the probe call

jaydeokar:
Can readiness also be hooked into the recovery path of programs ?

yash97:
> Can readiness also be hooked into the recovery path of programs ?

It can. But not scoping for this PR. Can do a followup for that.

pkg/rpc/healthcheck.go
yash97: Yes, there is a measurable cost. Benchmarked per-call `grpc.NewClient` vs. a shared client over the unix socket (grpc v1.79.3):

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| NewClient per call | ~432µs | ~184KB | 844 |
| shared client | ~59µs | ~9.8KB | 162 |

~7x slower and ~19x more allocations per probe. In absolute terms it is small at a 10s probe period, but `grpc.ClientConn` is meant to be created once and reused, so I refactored to a `GRPCSocketHealthChecker` that holds one client and is shared between `/healthz` and `/readyz`. Kept the per-call timeout, and `grpc.NewClient` is lazy so the "server not up yet" case still works.

main.go
yash97: Done, simplified to the single slice literal.

main.go
yash97: You are right. The events binary creates the `global_policy_events` map whenever it is missing, independent of `EnablePolicyEventLogs` (that flag only gates whether logging is configured). Verified on a cluster running with `--enable-policy-event-logs=false`: `/readyz?verbose` reports `[+]bpf-maps ok`. Made the readyz path unconditional.

pkg/ebpf/healthcheck_test.go
yash97: Good catch, it was not, it skipped on failure and asserted nothing on success. Removed it. The remaining two cases (non-existent path, and a regular dir that is not bpffs) cover the real failure modes.

pkg/rpc/healthcheck.go
viveksb007: i don't see the refactoring change, yet to push the commit?

yash97: yeah

pkg/rpc/healthcheck.go
```diff
@@ -0,0 +1,87 @@
...
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+)
+
+const grpcCheckTimeout = 10 * time.Second
+
+// GRPCSocketHealthChecker issues gRPC Health/Check RPCs against the NPA Unix
+// socket. It holds a single long-lived gRPC client reused across probe
+// invocations rather than rebuilding one per call.
+type GRPCSocketHealthChecker struct {
+	conn   *grpc.ClientConn
```
viveksb007, line 34: what is this `conn` supposed to do? we are not using this in `Check()` but some derived client

viveksb007: seems like its supposed to be a virutalConn https://pkg.go.dev/google.golang.org/grpc#ClientConn

yash97: yes we can manage state of connections with it.

viveksb007 (approved):
lgtm, Thanks for the changes

PR #576, opened 2026-05-15
fix(utils): make GetPodIdentifier injective to prevent cross-namespac…

parlakisik:
…e collisions

fix(utils): make GetPodIdentifier injective to prevent cross-namespace collisions

The per-pod BPF pin path key was "<stripped-pod-name>-<namespace>". Since
"-" is legal in both DNS-1123 pod names and namespaces, the boundary was
ambiguous: "deny-me-x"/"ns1" and "deny"/"me-ns1" both produced "deny-me-ns1",
so distinct pods shared programs and maps on a node. Switch the separator
to "@" (illegal in DNS-1123, accepted by bpffs, transparent to the SDK's
"_"-split pin parser).

On first startup, migrateLegacyPinsFromCNIState reads VPC CNI's
/var/run/aws-node/ipam.json and renames pin families <legacy>_* -> <new>_*
for each local pod. A marker at /var/run/aws-node/.npa_format_v2 makes the
migration a no-op on subsequent restarts.

Tests:
- Upgrade v1.3.1 → patched build: migration renamed all 16 legacy - pin families to @ format, marker written, agent log shows clean migration
- Cross-namespace identifier injectivity: GetPodIdentifier("deny-me-x", "ns1") and GetPodIdentifier("deny", "me-ns1") now produce distinct strings (deny-me@ns1 vs deny@me-ns1)
- unit tests

### DIFF
--- a/pkg/ebpf/legacy_pin_migration.go
+++ b/pkg/ebpf/legacy_pin_migration.go
@@ -1,0 +1,158 @@
+package ebpf
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+
+	"github.com/aws/aws-network-policy-agent/pkg/utils"
+)
+
+// cniIpamStatePath is VPC CNI's per-pod allocation checkpoint; already on a
+// path the NPA container mounts.
+const cniIpamStatePath = "/var/run/aws-node/ipam.json"
+
+// formatV2MarkerPath is the sentinel file written after a successful one-shot
+// legacy-format migration. Its presence makes the migration a no-op on every
+// subsequent agent restart.
+const formatV2MarkerPath = "/var/run/aws-node/.npa_format_v2"
+
+type ipamAllocation struct {
+	Metadata struct {
+		K8sPodName      string `json:"k8sPodName"`
+		K8sPodNamespace string `json:"k8sPodNamespace"`
+	} `json:"metadata"`
+}
+
+type ipamStateFile struct {
+	Version     string           `json:"version"`
+	Allocations []ipamAllocation `json:"allocations"`
+}
+
+// migrateLegacyPinsFromCNIState renames per-pod bpffs pin files from the
+// legacy "-" separator format to the new "_" format. Runs once per node;
+// the markerPath sentinel makes it a no-op on subsequent restarts.
+func migrateLegacyPinsFromCNIState(progsDir, mapsDir, ipamPath, markerPath string) error {
+	if _, err := os.Stat(markerPath); err == nil {
+		return nil
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("checking migration marker %s: %w", markerPath, err)
+	}
+
+	raw, err := os.ReadFile(ipamPath)
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("read CNI ipam state %s: %w", ipamPath, err)
+	}
+
+	var state ipamStateFile
+	if err := json.Unmarshal(raw, &state); err != nil {
+		return fmt.Errorf("parse CNI ipam state %s: %w", ipamPath, err)
+	}
+
+	byLegacyID := map[string][]ipamAllocation{}
+	for _, a := range state.Allocations {
+		if a.Metadata.K8sPodName == "" || a.Metadata.K8sPodNamespace == "" {
+			continue
+		}
+		legacyID := utils.LegacyGetPodIdentifier(a.Metadata.K8sPodName, a.Metadata.K8sPodNamespace)
+		byLegacyID[legacyID] = append(byLegacyID[legacyID], a)
+	}
+
+	var totalFailed int
+	for legacyID, pods := range byLegacyID {
+		// Sort by pod name so the choice of inheriting pod is deterministic.
+		// For multi-replica workloads all pods produce the same new ID, so
+		// order is irrelevant. For true cross-namespace collisions the first
+		// pod's pin gets renamed; the other pods will get fresh per-pod pins
+		// via the reconcile loop's normal attach path.
+		sort.Slice(pods, func(i, j int) bool {
+			return pods[i].Metadata.K8sPodName < pods[j].Metadata.K8sPodName
+		})
+		p := pods[0]
+		newID := utils.GetPodIdentifier(p.Metadata.K8sPodName, p.Metadata.K8sPodNamespace)
+
+		if len(pods) > 1 {
+			log().Infof("legacy pin %s shared by %d local pods; inheriting to first pod %s/%s (%s); other pods get fresh pins via reconcile",
+				legacyID, len(pods), p.Metadata.K8sPodNamespace, p.Metadata.K8sPodName, newID)
+		}
+		renamed, failed := renamePinFamily(progsDir, mapsDir, legacyID, newID)
+		totalFailed += failed
+		if renamed > 0 {
+			log().Infof("migrated %d bpffs pin file(s) for %s/%s: %s -> %s",
+				renamed, p.Metadata.K8sPodNamespace, p.Metadata.K8sPodName, legacyID, newID)
+		}
+	}
+
+	if totalFailed > 0 {
+		return fmt.Errorf("legacy pin migration incomplete: %d rename(s) failed; will retry on next restart", totalFailed)
+	}
+
+	if err := os.MkdirAll(filepath.Dir(markerPath), 0755); err != nil {
+		return fmt.Errorf("ensure marker dir for %s: %w", markerPath, err)
+	}
+	if err := os.WriteFile(markerPath, []byte("v2\n"), 0644); err != nil {
+		return fmt.Errorf("write migration marker %s: %w", markerPath, err)
+	}
+	log().Infof("legacy pin migration complete; marker written at %s", markerPath)
+	return nil
+}
+
+func renamePinFamily(progsDir, mapsDir, legacyID, newID string) (renamed int, failed int) {
+	for _, dir := range []string{"ingress", "egress"} {
+		progName := utils.TC_INGRESS_PROG
+		if dir == "egress" {
+			progName = utils.TC_EGRESS_PROG
+		}
+		switch renamePinIfExists(progsDir+legacyID+"_"+progName, progsDir+newID+"_"+progName) {
+		case renameOK:
+			renamed++
+		case renameFailed:
+			failed++
+		}
+	}
+	for _, mapName := range []string{
+		utils.TC_INGRESS_MAP,
+		utils.TC_EGRESS_MAP,
+		utils.TC_CLUSTER_POLICY_INGRESS_MAP,
+		utils.TC_CLUSTER_POLICY_EGRESS_MAP,
+		utils.TC_INGRESS_POD_STATE_MAP,
+		utils.TC_EGRESS_POD_STATE_MAP,
+	} {
+		switch renamePinIfExists(mapsDir+legacyID+"_"+mapName, mapsDir+newID+"_"+mapName) {
+		case renameOK:
+			renamed++
+		case renameFailed:
+			failed++
+		}
+	}
+	return renamed, failed
+}
+
+type renameResult int
+
+const (
+	renameSkipped renameResult = iota // source does not exist
+	renameOK                          // rename succeeded
+	renameFailed                      // source exists but rename failed
+)
+
+func renamePinIfExists(src, dst string) renameResult {
+	if _, err := os.Stat(src); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			log().Warnf("legacy pin migration: stat %s: %v", src, err)
+			return renameFailed
+		}
+		return renameSkipped
+	}
+	if err := os.Rename(src, dst); err != nil {
+		log().Warnf("legacy pin migration: rename %s -> %s: %v", src, dst, err)
+		return renameFailed
+	}
+	return renameOK
+}
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -169,13 +169,34 @@ func GetPodNamespacedName(podName, podNamespace string) string {
 	return podName + "_" + podNamespace
 }
 
+// Separator is "@" because it is illegal in DNS-1123 pod names and namespaces,
+// making the identifier injective on (podName-prefix, podNamespace). "@" also
+// keeps pin filenames parseable by aws-ebpf-sdk-go, which splits on the first
+// "_" to find the suffix boundary; using "_" here would shadow that split.
 func GetPodIdentifier(podName, podNamespace string) string {
 	if strings.Contains(podName, ".") {
 		log().Debug("Replacing '.' character with '_' for pod pin path.")
 		podName = strings.Replace(podName, ".", "_", -1)
 	}
 	podIdentifierPrefix := podName
-	if strings.Contains(string(podName), "-") {
+	if strings.Contains(podName, "-") {
+		tmpName := strings.Split(podName, "-")
+		podIdentifierPrefix = strings.Join(tmpName[:len(tmpName)-1], "-")
+	}
+	return podIdentifierPrefix + "@" + podNamespace
+}
+
+// LegacyGetPodIdentifier replicates the pre-fix GetPodIdentifier algorithm
+// with the "-" separator. Used only by the legacy bpffs pin migration to
+// compute what a previously-running agent would have produced for a given
+// (podName, podNamespace) pair.
+// TODO: remove after the migration window closes.
+func LegacyGetPodIdentifier(podName, podNamespace string) string {
+	if strings.Contains(podName, ".") {
+		podName = strings.Replace(podName, ".", "_", -1)
+	}
+	podIdentifierPrefix := podName
+	if strings.Contains(podName, "-") {
 		tmpName := strings.Split(podName, "-")
 		podIdentifierPrefix = strings.Join(tmpName[:len(tmpName)-1], "-")
 	}
--- a/controllers/policyendpoints_controller_test.go
+++ b/controllers/policyendpoints_controller_test.go
@@ -79,9 +79,9 @@ func TestPolicyEndpointReconcile(t *testing.T) {
 		assert.Nil(t, err)
 		val, ok := policyEndpointReconciler.networkPolicyToPodIdentifierMap.Load("allow-all-egress")
 		assert.True(t, ok)
-		assert.True(t, lo.Contains(val.([]string), "deployment1rs-my-namespace"))
+		assert.True(t, lo.Contains(val.([]string), "deployment1rs@my-namespace"))
 
-		val, ok = policyEndpointReconciler.podIdentifierToPolicyEndpointMap.Load("deployment1rs-my-namespace")
+		val, ok = policyEndpointReconciler.podIdentifierToPolicyEndpointMap.Load("deployment1rs@my-namespace")
 		assert.True(t, ok)
 		assert.True(t, lo.Contains(val.([]string), "allow-all-egress-abcd"))
 
@@ -125,9 +125,9 @@ func TestPolicyEndpointReconcile(t *testing.T) {
 		assert.Nil(t, err)
 		val, ok := policyEndpointReconciler.networkPolicyToPodIdentifierMap.Load("allow-all-egress")
 		assert.True(t, ok)
-		assert.True(t, lo.Contains(val.([]string), "deployment1rs-my-namespace"))
+		assert.True(t, lo.Contains(val.([]string), "deployment1rs@my-namespace"))
 
-		val, ok = policyEndpointReconciler.podIdentifierToPolicyEndpointMap.Load("deployment1rs-my-namespace")
+		val, ok = policyEndpointReconciler.podIdentifierToPolicyEndpointMap.Load("deployment1rs@my-namespace")
 		assert.True(t, ok)
 		assert.True(t, lo.Contains(val.([]string), "allow-all-egress-abcd"))
 
--- a/pkg/ebpf/bpf_client.go
+++ b/pkg/ebpf/bpf_client.go
@@ -392,6 +392,18 @@ func (l *bpfClient) recoverBPFState(bpfTCClient tc.BpfTc, eBPFSDKClient goelf.Bp
 	var interfaceNametoIngressPinPath = make(map[string]string)
 	var interfaceNametoEgressPinPath = make(map[string]string)
 
+	// Rename legacy "-" separator pin files to the new "_" format using
+	// VPC CNI's local pod inventory. Runs once per node, guarded by
+	// formatV2MarkerPath.
+	if err := migrateLegacyPinsFromCNIState(
+		utils.BPF_PROGRAMS_PIN_PATH_DIRECTORY,
+		utils.BPF_MAPS_PIN_PATH_DIRECTORY,
+		cniIpamStatePath,
+		formatV2MarkerPath,
+	); err != nil {
+		log().Errorf("legacy pin migration failed (non-fatal): %v", err)
+	}
+
 	// Recover global maps (Conntrack and Events) if there is no need to update
 	// events binary
 	if !updateEventsProbe {
@@ -789,7 +801,15 @@ func (l *bpfClient) attachIngressBPFProbe(hostVethName string, podIdentifier str
 
 	log().Infof("Attempting to do an Ingress Attach with progFD: %d", progFD)
 	err = l.bpfTCClient.TCEgressAttach(hostVethName, progFD, utils.TC_INGRESS_PROG)
-	if err != nil && !utils.IsFileExistsError(err.Error()) {
+	if err != nil && utils.IsFileExistsError(err.Error()) {
+		log().Warnf("Stale ingress TC filter on %s; detaching before re-attach", hostVethName)
+		if detachErr := l.bpfTCClient.TCEgressDetach(hostVethName); detachErr != nil {
+			log().Errorf("Ingress TC detach failed on %s: %v", hostVethName, detachErr)
+			return 0, detachErr
+		}
+		err = l.bpfTCClient.TCEgressAttach(hostVethName, progFD, utils.TC_INGRESS_PROG)
+	}
+	if err != nil {
 		log().Errorf("Ingress Attach failed %v", err)
 		return 0, err
 	}
@@ -826,7 +846,15 @@ func (l *bpfClient) attachEgressBPFProbe(hostVethName string, podIdentifier stri
 
 	log().Infof("Attempting to do an Egress Attach with progFD: %d", progFD)
 	err = l.bpfTCClient.TCIngressAttach(hostVethName, progFD, utils.TC_EGRESS_PROG)
-	if err != nil && !utils.IsFileExistsError(err.Error()) {
+	if err != nil && utils.IsFileExistsError(err.Error()) {
+		log().Warnf("Stale egress TC filter on %s; detaching before re-attach", hostVethName)
+		if detachErr := l.bpfTCClient.TCIngressDetach(hostVethName); detachErr != nil {
+			log().Errorf("Egress TC detach failed on %s: %v", hostVethName, detachErr)
+			return 0, detachErr
+		}
+		err = l.bpfTCClient.TCIngressAttach(hostVethName, progFD, utils.TC_EGRESS_PROG)
+	}
+	if err != nil {
 		log().Errorf("Egress Attach failed %v", err)
 		return 0, err
 	}
--- a/pkg/ebpf/c/tc.v4egress.bpf.c
+++ b/pkg/ebpf/c/tc.v4egress.bpf.c
@@ -482,4 +482,5 @@ int handle_egress(struct __sk_buff *skb)
 	return BPF_OK;
 }
 
+const volatile __u32 NPA_FILE_VERSION = 2;
 char _license[] SEC("license") = "GPL";
--- a/pkg/ebpf/c/tc.v4ingress.bpf.c
+++ b/pkg/ebpf/c/tc.v4ingress.bpf.c
@@ -465,4 +465,5 @@ int handle_ingress(struct __sk_buff *skb)
 	return BPF_OK;
 }
 
+const volatile __u32 NPA_FILE_VERSION = 2;
 char _license[] SEC("license") = "GPL";
--- a/pkg/ebpf/c/tc.v6egress.bpf.c
+++ b/pkg/ebpf/c/tc.v6egress.bpf.c
@@ -472,4 +472,5 @@ int handle_egress(struct __sk_buff *skb)
 	return BPF_OK;
 }
 
+const volatile __u32 NPA_FILE_VERSION = 2;
 char _license[] SEC("license") = "GPL";
--- a/pkg/ebpf/c/tc.v6ingress.bpf.c
+++ b/pkg/ebpf/c/tc.v6ingress.bpf.c
@@ -463,4 +463,5 @@ int handle_ingress(struct __sk_buff *skb)
 	return BPF_OK;
 }
 
+const volatile __u32 NPA_FILE_VERSION = 2;
 char _license[] SEC("license") = "GPL";
--- a/pkg/ebpf/legacy_pin_migration_test.go
+++ b/pkg/ebpf/legacy_pin_migration_test.go
@@ -1,0 +1,220 @@
+package ebpf
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/aws/aws-network-policy-agent/pkg/utils"
+	"github.com/stretchr/testify/assert"
+)
+
+// writeIpam writes a minimal ipam.json with the given (podName, podNamespace) pairs.
+func writeIpam(t *testing.T, dir string, pods [][2]string) string {
+	t.Helper()
+	allocs := ""
+	for i, p := range pods {
+		if i > 0 {
+			allocs += ","
+		}
+		allocs += `{"metadata":{"k8sPodName":"` + p[0] + `","k8sPodNamespace":"` + p[1] + `"}}`
+	}
+	path := filepath.Join(dir, "ipam.json")
+	body := `{"version":"vpc-cni-ipam/1","allocations":[` + allocs + `]}`
+	if err := os.WriteFile(path, []byte(body), 0644); err != nil {
+		t.Fatalf("write ipam: %v", err)
+	}
+	return path
+}
+
+func TestMigrateLegacyPinsFromCNIState_HappyPath(t *testing.T) {
+	progsDir := t.TempDir() + "/"
+	mapsDir := t.TempDir() + "/"
+	ipamPath := writeIpam(t, t.TempDir(), [][2]string{
+		{"deny-me-0", "ns1"},
+		{"web-frontend-0", "my-namespace"},
+		// Namespace containing "-" — boundary must be the prefix/ns join,
+		// the "-" inside the namespace must be PRESERVED.
+		{"aws-node-xyz", "kube-system"},
+	})
+
+	legacyFiles := []struct{ dir, name string }{
+		{progsDir, "deny-me-ns1_" + utils.TC_INGRESS_PROG},
+		{progsDir, "deny-me-ns1_" + utils.TC_EGRESS_PROG},
+		{mapsDir, "deny-me-ns1_" + utils.TC_INGRESS_MAP},
+		{mapsDir, "deny-me-ns1_" + utils.TC_EGRESS_POD_STATE_MAP},
+		{progsDir, "web-frontend-my-namespace_" + utils.TC_INGRESS_PROG},
+		{mapsDir, "web-frontend-my-namespace_" + utils.TC_CLUSTER_POLICY_INGRESS_MAP},
+		{progsDir, "aws-node-kube-system_" + utils.TC_INGRESS_PROG},
+		{mapsDir, "aws-node-kube-system_" + utils.TC_INGRESS_MAP},
+	}
+	for _, f := range legacyFiles {
+		assert.NoError(t, os.WriteFile(f.dir+f.name, []byte("x"), 0644))
+	}
+
+	if err := migrateLegacyPinsFromCNIState(progsDir, mapsDir, ipamPath, filepath.Join(t.TempDir(), ".npa_format_v2")); err != nil {
+		t.Fatalf("migrate: %v", err)
+	}
+
+	expectations := map[string]bool{
+		// deny-me/ns1
+		progsDir + "deny-me-ns1_" + utils.TC_INGRESS_PROG:        false,
+		progsDir + "deny-me@ns1_" + utils.TC_INGRESS_PROG:        true,
+		progsDir + "deny-me@ns1_" + utils.TC_EGRESS_PROG:         true,
+		mapsDir + "deny-me@ns1_" + utils.TC_INGRESS_MAP:          true,
+		mapsDir + "deny-me@ns1_" + utils.TC_EGRESS_POD_STATE_MAP: true,
+		// web-frontend/my-namespace
+		progsDir + "web-frontend@my-namespace_" + utils.TC_INGRESS_PROG:              true,
+		mapsDir + "web-frontend@my-namespace_" + utils.TC_CLUSTER_POLICY_INGRESS_MAP: true,
+		progsDir + "web-frontend-my-namespace_" + utils.TC_INGRESS_PROG:              false,
+		// aws-node/kube-system — boundary must become "@" but "kube-system" stays intact
+		progsDir + "aws-node-kube-system_" + utils.TC_INGRESS_PROG: false,
+		progsDir + "aws-node@kube-system_" + utils.TC_INGRESS_PROG: true,
+		mapsDir + "aws-node@kube-system_" + utils.TC_INGRESS_MAP:   true,
+		// Anti-assertion: must NOT corrupt "kube-system" by changing its internal "-"
+		progsDir + "aws-node@kube_system_" + utils.TC_INGRESS_PROG: false,
+		mapsDir + "aws-node@kube_system_" + utils.TC_INGRESS_MAP:   false,
+	}
+	for path, shouldExist := range expectations {
+		_, err := os.Stat(path)
+		if shouldExist {
+			assert.NoError(t, err, "expected %s to exist", path)
+		} else {
+			assert.True(t, os.IsNotExist(err), "expected %s to be gone", path)
+		}
+	}
+}
+
+// When two pods share a legacy identifier (the cross-namespace bypass case),
+// the legacy pin is renamed to the first pod's new identifier (sorted by name).
+// The other pod will get a fresh per-pod pin via the reconcile loop.
+func TestMigrateLegacyPinsFromCNIState_CollisionRenamesToFirstPod(t *testing.T) {
+	progsDir := t.TempDir() + "/"
+	mapsDir := t.TempDir() + "/"
+	ipamPath := writeIpam(t, t.TempDir(), [][2]string{
+		{"deny-me-x", "ns1"}, // legacy: deny-me-ns1, new: deny-me@ns1
+		{"deny-x", "me-ns1"}, // legacy: deny-me-ns1, new: deny@me-ns1
+	})
+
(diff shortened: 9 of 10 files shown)


viveksb007 (commented):
is there actual upgrade/downgrade test done for this change?

like cx will be on older image where PodId will be different and upgrading will gracefully move to new PodId and will downgrading also be graceful?

pkg/utils/utils.go
```diff
@@ -153,17 +153,33 @@ func GetPodNamespacedName(podName, podNamespace string) string {
...
+// the final separator. Returns the input unchanged when already new format.
+// TODO: remove after the migration window closes.
+func TranslateLegacyPodIdentifier(id string) string {
+	lastDash := strings.LastIndex(id, "-")
+	lastUnderscore := strings.LastIndex(id, "_")
+	if lastDash <= lastUnderscore {
+		return id
+	}
+	return id[:lastDash] + "_" + id[lastDash+1:]
 }
```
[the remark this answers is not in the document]
Pavani-Panakanti: having '-' in namespaces is very common and it will break those. We can't use this approach for conversion. looks like we have to know pod names and namespaces and cannot modify from podidentifier

Pavani-Panakanti:
Looks like downgrades won't work. We will have old bpf programs and maps from new version left over. We should see how we want to handle these and test them as well

parlakisik:
We can not find solution for downgrades. customer can upgrade any version they want .
this change is not backward compatible.

parlakisik:
Testing :

- Pre-state on NPA v1.3.1: deployed deny-me-0 + deny-all-ingress NetworkPolicy, confirmed bpffs had legacy <prefix>-<ns>_* pins and no marker, probe was BLOCKED.
- Upgrade to new Fixed image: rolled the aws-eks-nodeagent daemonset to the build under test and waited for rollout completion.
- Post-upgrade migration check: verified legacy --format pins were wiped, the .npa_format_v2 marker was written at /var/run/aws-node/, and the two global maps (global_aws_conntrack_map, global_policy_events) survived.
- Existing-pod enforcement: re-probed deny-me-0 after the wipe and confirmed it stayed BLOCKED — the kernel TC filter kept the program live via FD ref even with the pin gone.
- Re-attach with new format: rolled-restarted the workload and confirmed the new pod's pins now use _ separator (deny-me_ns1_*) with no stale legacy entries.
- New post-upgrade workload: created a fresh namespace + pod + NetworkPolicy on the upgraded agent, confirmed enforcement (BLOCKED) and new-format pins, plus 0 panic/fatal lines across NPA logs on all nodes.

pkg/ebpf/legacy_pin_migration.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+		legacyID := utils.LegacyGetPodIdentifier(a.Metadata.K8sPodName, a.Metadata.K8sPodNamespace)
+		byLegacyID[legacyID] = append(byLegacyID[legacyID], a)
+	}
+
+	for legacyID, pods := range byLegacyID {
+		// Sort by pod name so the choice of inheriting pod is deterministic.
+		// For multi-replica workloads all pods produce the same new ID, so
+		// order is irrelevant. For true cross-namespace collisions the first
+		// pod's pin gets renamed; the other pods will get fresh per-pod pins
```
viveksb007, line 72: I think this assumption is unsafe for the actual upgrade path.

In a collision case, two local pods can have different `(podName, namespace)` values but the same legacy pod identifier. For example:

```text
deny-me-x / ns1     -> legacy: deny-me-ns1, new: deny-me@ns1
deny-x    / me-ns1  -> legacy: deny-me-ns1, new: deny@me-ns1
```

The migration renames the legacy bpffs pins to the first pod's new identifier. That moves the pinned program/map names, but it does not change any TC filters already attached to the pods' veth interfaces. Any colliding pod that does not inherit the renamed pin may still have its TC filter attached to the old shared program FD.

The later reconcile path may not fix that. When `attachIngressBPFProbe` / `attachEgressBPFProbe` attempts to attach a new program to the non-inheriting pod's existing interface, `FilterAdd` can fail with a file-exists error because a filter is already attached. The current attach code treats that error as success:

```go
if err != nil && !utils.IsFileExistsError(err.Error()) {
    return 0, err
}
return progFD, nil
```

So the agent can record the new program in memory while the kernel TC filter remains attached to the old shared program, now associated with the inheriting pod's maps. That means the non-inheriting pod can continue being enforced with the wrong policy state after upgrade.

I think collision migration needs to explicitly handle attached filters for all colliding pods, not only rename pin files. Possible fixes are:

```text
1. In true collision cases, do not inherit the shared legacy program for only one pod; force all colliding pods through clean detach/recreate.
2. Or, for non-inheriting pods, explicitly detach the existing TC filters before reconcile/attach.
3. Or, record interfaces for colliding pods and force replace rather than accepting file-exists as success.
```

A regression test would be useful for the case where two pods share a legacy identifier and one of them already has an attached TC filter.

pkg/utils/utils.go
```diff
@@ -157,13 +157,37 @@ func GetPodNamespacedName(podName, podNamespace string) string {
...
+// keeps pin filenames parseable by aws-ebpf-sdk-go, which splits on the first
+// "_" to find the suffix boundary; using "_" here would shadow that split.
+// "." is unusable because the kernel BPF filesystem rejects rename to filenames
+// containing ".", which would block the legacy-pin migration.
 func GetPodIdentifier(podName, podNamespace string) string {
 	if strings.Contains(podName, ".") {
-		log().Debug("Replacing '.' character with '_' for pod pin path.")
+		// Substitute with "-" (DNS-1123-safe and not our boundary character)
+		// so the joined identifier never contains "." outside the boundary.
+		podName = strings.Replace(podName, ".", "-", -1)
```
viveksb007: This makes the new identifier non-injective for valid Kubernetes pod names. `.` and `-` are both valid in pod names, so replacing `.` with `-` folds distinct pod names into the same value before the replica suffix is trimmed.

Example:

```text
my.pod-x / ns -> my-pod-x -> prefix my-pod -> my-pod@ns
my-pod-x / ns -> my-pod-x -> prefix my-pod -> my-pod@ns
```

These are distinct valid pods, but they produce the same new pod identifier. Since this identifier is used for BPF program/map sharing and policy cache lookup, this can merge enforcement state for unrelated pods.

I understand why `_` is problematic here: aws-ebpf-sdk-go parses map pin names using the first `_`, so putting `_` inside the customized namespace/prefix breaks recovery for namespaced maps. But replacing `.` with `-` is still lossy.

Can we switch this to a reversible/bijective escape that is safe for bpffs and does not use `_` or `.`? For example, encode special characters using a sentinel that cannot appear unescaped, such as a `~`-based escape, assuming `~` is accepted by bpffs and by the SDK parsing path:

```text
"." -> "~d"
"~" -> "~t"
```

The exact escape scheme can vary, but the important property is that two different valid pod names cannot normalize to the same pod identifier. I would also add a unit test like:

```go
assert.NotEqual(
    GetPodIdentifier("my.pod-x", "ns"),
    GetPodIdentifier("my-pod-x", "ns"),
)
```

pkg/ebpf/legacy_pin_migration.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func renamePinIfExists(src, dst string) bool {
+	if _, err := os.Stat(src); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			log().Warnf("legacy pin migration: stat %s: %v", src, err)
+		}
+		return false
+	}
+	if err := os.Rename(src, dst); err != nil {
```
viveksb007, line 153: `renamePinIfExists` logs failures and returns `false`, but `migrateLegacyPinsFromCNIState` still writes the one-shot marker at the end:

```go
if err := os.WriteFile(markerPath, []byte("v2\n"), 0644); err != nil {
    return fmt.Errorf("write migration marker %s: %w", markerPath, err)
}
```

That means a transient failure during one or more pin renames permanently suppresses future migration attempts. For example, if a rename fails due to a temporary bpffs issue, permission/mount timing issue, or destination conflict, the marker is still written and those legacy pins will never be migrated on subsequent agent restarts.

This seems risky because the rest of the PR relies on the marker being a signal that migration completed successfully, not merely that migration was attempted. If any legacy pins are left behind, recovery can still see old identifiers or skip expected new identifiers, depending on which files failed to rename.

I suggest making rename failures visible to the top-level migration result and only writing the marker after all expected renames either succeed or are confirmed unnecessary. For example:

```text
- return `(renamed int, err error)` from renamePinFamily / renamePinIfExists
- aggregate errors across the family
- skip writing `.npa_format_v2` if any rename failed
- allow the next restart to retry
```

It would also be useful to add a test where `dst` already exists or `os.Rename` fails, and assert that the marker is not written.

parlakisik: we are trying to detach once  if it doesnt solve the issue . then create error

pkg/utils/utils.go
parlakisik: this part is left over from previous code removed .

pkg/ebpf/legacy_pin_migration.go
parlakisik: resolved

pkg/ebpf/legacy_pin_migration.go
parlakisik: On upgrade the eBPF binary changes ,agent detects the new binary and wipes all TC filters then re-attaches every pod with the new program from scratch. The pin rename (migration) just ensures the recovered program context uses the correct new identifier so each pod gets its own independent maps.

parlakisik:
RESULTS
  PRE-upgrade legacy pins seen                 : 8    (expect > 0)
  POST-upgrade legacy pins remaining           : 0    (expect 0)
  POST-upgrade new-format pins (deny-me@ns1_*) : 8     (expect > 0)
  POST-upgrade .npa_format_v2 marker           : 1 (expect >= 1)
  POST-upgrade global_aws_conntrack_map        : 1 (expect 1)
  POST-upgrade global_policy_events            : 1    (expect 1)
  Existing pods BLOCKED after upgrade          : 1/1
  POST-restart new-format pins                 : 8    (expect > 0)
  POST-restart stale legacy pins               : 0  (expect 0)
  Pods BLOCKED after rollout restart           : 1/1
  Replicas (shared podId) all BLOCKED          : 1/1
  Cross-ns collision — victim pins             : 8
  Cross-ns collision — attacker pins           : 8
  Collision bug visible pre-upgrade              : 1/1
  Collision fix confirmed post-upgrade          : 1/1
  Policy mutation — victim allow               : 1/1
  Policy mutation — victim deny restored       : 1/1
  Policy mutation — no crosstalk to attacker   : 1/1
  FRESH ns+pod+policy BLOCKED                  : 1/1
  FRESH pod new-format pins                    : 8 (expect > 0)
  FRESH pod legacy pins                        : 0    (expect 0)
  Agent restart idempotency — new-format pins  : 8 (expect > 0)
  Agent restart idempotency — legacy pins      : 0     (expect 0)
  Agent restart — enforcement maintained       : 1/1
  Agent container restarts                     : 0     (expect 0)
  Migration log seen in agent output           : 0  (expect >= 1)
  NPA log panic/fatal/map-error lines          : 0         (expect 0)

parlakisik:
<h2 style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Upgrade Validation</h2><h3 style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Setup</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Two StatefulSets deployed on the same node to trigger the cross-namespace collision:</p><ul style="padding-inline-start: 2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li><strong><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">deny-me</code><span> </span>StatefulSet</strong><span> </span>in namespace<span> </span><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">ns1</code><span> </span>(port 80) — has a NetworkPolicy allowing only TCP:80 from same namespace + egress to DNS</li><li><strong><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">deny</code><span> </span>StatefulSet</strong><span> </span>in namespace<span> </span><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">me-ns1</code><span> </span>(port 8080) — has<span> </span><strong>no</strong><span> </span>NetworkPolicy (should be open)</li><li><strong><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">probe</code><span> </span>Pod</strong><span> </span>in namespace<span> </span><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">probe</code><span> </span>— connectivity tester (busybox, no policy)</li></ul><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Under the old agent (v1.3.1), both StatefulSets produce BPF identifier <code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">deny-me-ns1</code> and share a single program + maps on the node.</p><h3 style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">What was validated</h3>
# | Test | Result


1 | Collision bug exists pre-upgrade — no-policy pod is incorrectly BLOCKED by victim's rules 

2 | Migration renames pins — legacy deny-me-ns1_* → distinct deny-me@ns1_* + deny@me-ns1_* 

3 | Collision resolved post-upgrade — no-policy pod becomes REACHABLE 

4 | Existing enforcement survives upgrade — victim stays BLOCKED 

5 | Replica sharing still works — scaled to 3 replicas, all BLOCKED (shared podIdentifier) 

6 | No stale pins after rollout restart — only new-format pins remain 

7 | No policy crosstalk — mutating victim's policy doesn't affect attacker pod 

8 | Fresh pods work — new namespace + pod + policy enforced correctly post-upgrade 

9 | Agent restart idempotent — marker prevents re-migration, 0 container restarts 

10 | Zero panics/errors in agent logs 


<h3 style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">BPF state before vs after</h3><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>Before upgrade (v1.3.1):</strong></p><div class="codeBlockWrapper_-a7MRw" style="position: relative; margin: 8px 0px; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><button class="copyButton_CEmTFw copyButton_-a7MRw" title="Copy code" aria-label="Copy code to clipboard" style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; background: none 0% 0% / auto repeat scroll padding-box border-box rgb(18, 19, 20); border-color: rgb(42, 43, 44); border-style: solid; border-width: 1px; border-image: none 100% / 1 / 0 stretch; cursor: pointer; opacity: 0; display: flex; border-radius: 4px; justify-content: center; align-items: center; padding: 4px; transition: opacity 0.15s, background 0.15s; position: absolute; top: 4px; right: 4px;"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true" data-slot="icon" class="copyIcon_CEmTFw"><path fill-rule="evenodd" d="M15.988 3.012A2.25 2.25 0 0 1 18 5.25v6.5A2.25 2.25 0 0 1 15.75 14H13.5v-3.379a3 3 0 0 0-.879-2.121l-3.12-3.121a3 3 0 0 0-1.402-.791 2.252 2.252 0 0 1 1.913-1.576A2.25 2.25 0 0 1 12.25 1h1.5a2.25 2.25 0 0 1 2.238 2.012ZM11.5 3.25a.75.75 0 0 1 .75-.75h1.5a.75.75 0 0 1 .75.75v.25h-3v-.25Z" clip-rule="evenodd"></path><path d="M3.5 6A1.5 1.5 0 0 0 2 7.5v9A1.5 1.5 0 0 0 3.5 18h7a1.5 1.5 0 0 0 1.5-1.5v-5.879a1.5 1.5 0 0 0-.44-1.06L8.44 6.439A1.5 1.5 0 0 0 7.378 6H3.5Z"></path></svg></button><pre style="overflow-x: auto; white-space: pre; box-sizing: border-box; border-radius: 4px; max-width: 100%; margin: 0px; padding: 8px;"><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 0px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">deny-me-ns1_handle_ingress    ← SHARED by both pods
deny-me-ns1_handle_egress
deny-me-ns1_ingress_map       ← last-write-wins: one pod's rules overwrite the other
deny-me-ns1_egress_map
...
(no marker)
</code></pre></div><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong>After upgrade:</strong></p><div class="codeBlockWrapper_-a7MRw" style="position: relative; margin: 8px 0px; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><button class="copyButton_CEmTFw copyButton_-a7MRw" title="Copy code" aria-label="Copy code to clipboard" style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; background: none 0% 0% / auto repeat scroll padding-box border-box rgb(18, 19, 20); border-color: rgb(42, 43, 44); border-style: solid; border-width: 1px; border-image: none 100% / 1 / 0 stretch; cursor: pointer; opacity: 0; display: flex; border-radius: 4px; justify-content: center; align-items: center; padding: 4px; transition: opacity 0.15s, background 0.15s; position: absolute; top: 4px; right: 4px;"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true" data-slot="icon" class="copyIcon_CEmTFw"><path fill-rule="evenodd" d="M15.988 3.012A2.25 2.25 0 0 1 18 5.25v6.5A2.25 2.25 0 0 1 15.75 14H13.5v-3.379a3 3 0 0 0-.879-2.121l-3.12-3.121a3 3 0 0 0-1.402-.791 2.252 2.252 0 0 1 1.913-1.576A2.25 2.25 0 0 1 12.25 1h1.5a2.25 2.25 0 0 1 2.238 2.012ZM11.5 3.25a.75.75 0 0 1 .75-.75h1.5a.75.75 0 0 1 .75.75v.25h-3v-.25Z" clip-rule="evenodd"></path><path d="M3.5 6A1.5 1.5 0 0 0 2 7.5v9A1.5 1.5 0 0 0 3.5 18h7a1.5 1.5 0 0 0 1.5-1.5v-5.879a1.5 1.5 0 0 0-.44-1.06L8.44 6.439A1.5 1.5 0 0 0 7.378 6H3.5Z"></path></svg></button><pre style="overflow-x: auto; white-space: pre; box-sizing: border-box; border-radius: 4px; max-width: 100%; margin: 0px; padding: 8px;"><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 0px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">deny-me@ns1_handle_ingress    ← victim's own program
deny-me@ns1_ingress_map       ← victim's own rules
deny@me-ns1_handle_ingress    ← attacker's own program (empty maps, no policy)
deny@me-ns1_ingress_map
...
/var/run/aws-node/.npa_format_v2  ← migration marker (prevents re-run)
</code></pre></div><h3 style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Migration log from node</h3><div class="codeBlockWrapper_-a7MRw" style="position: relative; margin: 8px 0px; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(25, 26, 27); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><button class="copyButton_CEmTFw copyButton_-a7MRw" title="Copy code" aria-label="Copy code to clipboard" style="color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; background: none 0% 0% / auto repeat scroll padding-box border-box rgb(18, 19, 20); border-color: rgb(42, 43, 44); border-style: solid; border-width: 1px; border-image: none 100% / 1 / 0 stretch; cursor: pointer; opacity: 0; display: flex; border-radius: 4px; justify-content: center; align-items: center; padding: 4px; transition: opacity 0.15s, background 0.15s; position: absolute; top: 4px; right: 4px;"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true" data-slot="icon" class="copyIcon_CEmTFw"><path fill-rule="evenodd" d="M15.988 3.012A2.25 2.25 0 0 1 18 5.25v6.5A2.25 2.25 0 0 1 15.75 14H13.5v-3.379a3 3 0 0 0-.879-2.121l-3.12-3.121a3 3 0 0 0-1.402-.791 2.252 2.252 0 0 1 1.913-1.576A2.25 2.25 0 0 1 12.25 1h1.5a2.25 2.25 0 0 1 2.238 2.012ZM11.5 3.25a.75.75 0 0 1 .75-.75h1.5a.75.75 0 0 1 .75.75v.25h-3v-.25Z" clip-rule="evenodd"></path><path d="M3.5 6A1.5 1.5 0 0 0 2 7.5v9A1.5 1.5 0 0 0 3.5 18h7a1.5 1.5 0 0 0 1.5-1.5v-5.879a1.5 1.5 0 0 0-.44-1.06L8.44 6.439A1.5 1.5 0 0 0 7.378 6H3.5Z"></path></svg></button><pre style="overflow-x: auto; white-space: pre; box-sizing: border-box; border-radius: 4px; max-width: 100%; margin: 0px; padding: 8px;"><code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 0px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">migrated 8 bpffs pin file(s) for me-ns1/deny-0: deny-me-ns1 -&gt; deny@me-ns1
migrated 8 bpffs pin file(s) for ns1/deny-me-0: deny-me-ns1 -&gt; deny-me@ns1
legacy pin migration complete; marker written at /var/run/aws-node/.npa_format_v2</code></pre></div>

pkg/ebpf/legacy_pin_migration.go
viveksb007: > On upgrade the eBPF binary changes

I don't see bpf binary change in this PR. will upgrading to version which has this PR change result in the action of removing all TC filters and re-attachment of probes?

can you check for case where - "there are 2 pods which have this colliding name and sharing TC filter and some non-colliding pods, what will be the PIN prog and map states (name) and TC filter ID", then once we upgrade cluster to this version, what will be the PIN prog and map states (name) and TC filter ID of all those pods.

parlakisik: in each build release epbf compiled and generating new id . so in each release system assumes that there is a new image

Pavani-Panakanti: It will compare new and old binaries. There is no id associated with ebpf binaries we ship with np image. If there is no change in any of these binaries it won't trigger binary change flow. We should make some change in bpf or events file if we want to trigger that flow. 
It would be good to evaluate with a test like Vivek pointed out

parlakisik:
After some tests 


Interface | Pre Ingress | Pre Egress | Post Ingress | Post Egress | Pod | Pin (post-upgrade)
-- | -- | -- | -- | -- | -- | --
eni22b168d2b96 | 237 | 235 | 360 | 361 | deny-me-0/ns1 (victim) | deny-me@ns1_handle_*
eni82da2737e58 | 237 | 235 | 332 | 333 | deny-0/me-ns1 (attacker) | deny@me-ns1_handle_*
eni9e9b97dc109 | 155 | 154 | 331 | 337 | coredns replica 1 | coredns-59c5997fcc@kube-system_handle_*
eniccfd55cfa49 | 155 | 154 | 331 | 337 | coredns replica 2 | coredns-59c5997fcc@kube-system_handle_*
eni06e4b6045b0 | 156 | 153 | 330 | 336 | metrics-server replica 1 | metrics-server-d994994bf@kube-system_handle_*
enif8cbf8d4e4e | 156 | 153 | 330 | 336 | metrics-server replica 2 | metrics-server-d994994bf@kube-system_handle_*
enie0227674628 | 142 | 141 | 329 | 335 | external-dns | external-dns-7547d8dd46@external-dns_handle_*
enia57dbd56ff2 | 243 | 236 | 328 | 334 | probe | probe@probe_handle_*

parlakisik:
Also same image is rebuild with different hash 

(26-06-05 22:42:56) <0> [~/eks/github/aws-network-policy-agent/upgrade-test]  
dev-dsk-parlmura-2b-4b813a33 % docker images 
REPOSITORY                                                              TAG                     IMAGE ID       CREATED             SIZE
aws-network-policy-agent   identifierupdate-same   41ccdb32c38d   About an hour ago   155MB
aws-network-policy-agent   identifierupdate-v3     91036e2acb9f   2 days ago          155MB

the upgrade changes the program ids 


Interface | v3 Ingress/Egress | same Ingress/Egress | Pod
-- | -- | -- | --
eni22b168d2b96 | 360/361 | 462/472 | deny-me-0/ns1
eni82da2737e58 | 332/333 | 466/470 | deny-0/me-ns1
eni9e9b97dc109 | 331/337 | 464/471 | coredns

<br class="Apple-interchange-newline">

```

{"level":"info","ts":"2026-06-05T22:42:10.020Z","caller":"ebpf/bpf_client.go:169","msg":"comparing new and existing probes ..."}
{"level":"info","ts":"2026-06-05T22:42:10.024Z","caller":"ebpf/bpf_client.go:169","msg":"change detected in event probe binaries.."}
{"level":"info","ts":"2026-06-05T22:42:10.044Z","caller":"ebpf/bpf_client.go:188","msg":"Collected all data for reattaching probes"}
{"level":"info","ts":"2026-06-05T22:42:10.391Z","caller":"ebpf/bpf_client.go:581","msg":"Attempting to do an Ingress Attach with progFD: 13"}
{"level":"info","ts":"2026-06-05T22:42:10.709Z","caller":"ebpf/bpf_client.go:581","msg":"Attempting to do an Ingress Attach with progFD: 20"}
{"level":"info","ts":"2026-06-05T22:42:11.051Z","caller":"ebpf/bpf_client.go:581","msg":"Attempting to do an Ingress Attach with progFD: 24"}
{"level":"info","ts":"2026-06-05T22:42:11.385Z","caller":"ebpf/bpf_client.go:581","msg":"Attempting to do an Ingress Attach with progFD: 28"}
{"level":"info","ts":"2026-06-05T22:42:11.730Z","caller":"ebpf/bpf_client.go:581","msg":"Attempting to do an Ingress Attach with progFD: 32"}
{"level":"info","ts":"2026-06-05T22:42:11.730Z","caller":"ebpf/bpf_client.go:581","msg":"Attempting to do an Ingress Attach with progFD: 28"}
{"level":"info","ts":"2026-06-05T22:42:12.132Z","caller":"ebpf/bpf_client.go:581","msg":"Attempting to do an Ingress Attach with progFD: 36"}
{"level":"info","ts":"2026-06-05T22:42:12.469Z","caller":"ebpf/bpf_client.go:581","msg":"Attempting to do an Ingress Attach with progFD: 40"}
{"level":"info","ts":"2026-06-05T22:42:12.469Z","caller":"ebpf/bpf_client.go:581","msg":"Attempting to do an Ingress Attach with progFD: 24"}
{"level":"info","ts":"2026-06-05T22:42:12.812Z","caller":"ebpf/bpf_client.go:601","msg":"Attempting to do an Egress Attach with progFD: 44"}
{"level":"info","ts":"2026-06-05T22:42:13.182Z","caller":"ebpf/bpf_client.go:601","msg":"Attempting to do an Egress Attach with progFD: 48"}
{"level":"info","ts":"2026-06-05T22:42:13.553Z","caller":"ebpf/bpf_client.go:601","msg":"Attempting to do an Egress Attach with progFD: 52"}
{"level":"info","ts":"2026-06-05T22:42:13.911Z","caller":"ebpf/bpf_client.go:601","msg":"Attempting to do an Egress Attach with progFD: 56"}
{"level":"info","ts":"2026-06-05T22:42:13.911Z","caller":"ebpf/bpf_client.go:601","msg":"Attempting to do an Egress Attach with progFD: 56"}
{"level":"info","ts":"2026-06-05T22:42:13.912Z","caller":"ebpf/bpf_client.go:601","msg":"Attempting to do an Egress Attach with progFD: 44"}
{"level":"info","ts":"2026-06-05T22:42:14.289Z","caller":"ebpf/bpf_client.go:601","msg":"Attempting to do an Egress Attach with progFD: 60"}
{"level":"info","ts":"2026-06-05T22:42:14.708Z","caller":"ebpf/bpf_client.go:601","msg":"Attempting to do an Egress Attach with progFD: 64"}
{"level":"info","ts":"2026-06-05T22:42:15.104Z","caller":"ebpf/bpf_client.go:601","msg":"Attempting to do an Egress Attach with progFD: 68"}
{"level":"info","ts":"2026-06-05T22:42:15.252Z","caller":"ebpf/bpf_client.go:736","msg":"Attempting to do an Ingress Attach with progFD: 40"}
{"level":"info","ts":"2026-06-05T22:42:15.253Z","caller":"ebpf/bpf_client.go:749","msg":"Attempting to do an Egress Attach with progFD: 68"}
{"level":"info","ts":"2026-06-05T22:42:15.255Z","caller":"ebpf/bpf_client.go:736","msg":"Attempting to do an Ingress Attach with progFD: 20"}
{"level":"info","ts":"2026-06-05T22:42:15.256Z","caller":"ebpf/bpf_client.go:749","msg":"Attempting to do an Egress Attach with progFD: 60"}
```

parlakisik:
FILE           identifierupdate                                            committed (NPA_FILE_VERSION=2)       
----           --------------------------------                         ------------------------------                  
tc.v4ingress   677aa7f7910172ed0c7a6168c85a5a2d   41de36298506fb1daf9e488f6926f72e                                 
tc.v4egress    dfe751318df4a1a7ac2100cbcf77bc8e   92f4fd18473e76da8a5357eba5ed6352                                 
tc.v6ingress   79814b84bafba53cf8b17b63b55c4e86   d321228ba0cd05460ad755a4a7a8f99e                                 
tc.v6egress    742db89f3d14d5cf4e09ca7b44371d1c   52847e0e649e6742d6a3857fd024cf1a                                 
v4events       a3951e5c23c106bf205a29bbb71ec575   dfdcd25c11fc7f4b4cf54b055c6d030d                                 
v6events       3201f9d9771ae1ed23c50f8c2f6472c9   3000ee477c0546ca3ab3f2079a6bf88d

pkg/ebpf/legacy_pin_migration.go
parlakisik: added ebpf file .

parlakisik:
```
parlmura@dev-dsk-parlmura-2b-4b813a33 aws-network-policy-agent]$ for o in $FILES; do md5sum /tmp/same_${o}.o; done
677aa7f7910172ed0c7a6168c85a5a2d  /tmp/same_tc.v4ingress.o
dfe751318df4a1a7ac2100cbcf77bc8e  /tmp/same_tc.v4egress.o
79814b84bafba53cf8b17b63b55c4e86  /tmp/same_tc.v6ingress.o
742db89f3d14d5cf4e09ca7b44371d1c  /tmp/same_tc.v6egress.o
a3951e5c23c106bf205a29bbb71ec575  /tmp/same_v4events.o
3201f9d9771ae1ed23c50f8c2f6472c9  /tmp/same_v6events.o
[parlmura@dev-dsk-parlmura-2b-4b813a33 aws-network-policy-agent]$  for o in $FILES; do md5sum /tmp/committed_${o}.o; done
41de36298506fb1daf9e488f6926f72e  /tmp/committed_tc.v4ingress.o
92f4fd18473e76da8a5357eba5ed6352  /tmp/committed_tc.v4egress.o
d321228ba0cd05460ad755a4a7a8f99e  /tmp/committed_tc.v6ingress.o
52847e0e649e6742d6a3857fd024cf1a  /tmp/committed_tc.v6egress.o
dfdcd25c11fc7f4b4cf54b055c6d030d  /tmp/committed_v4events.o
3000ee477c0546ca3ab3f2079a6bf88d  /tmp/committed_v6events.o
[parlmura@dev-dsk-parlmura-2b-4b813a33 aws-network-policy-agent]$  for o in $FILES; do echo "----- ${o}.bpf.o -----";  cmp /tmp/same_${o}.o /tmp/committed_${o}.o; 
> done
----- tc.v4ingress.bpf.o -----
/tmp/same_tc.v4ingress.o /tmp/committed_tc.v4ingress.o differ: byte 41, line 1
----- tc.v4egress.bpf.o -----
/tmp/same_tc.v4egress.o /tmp/committed_tc.v4egress.o differ: byte 41, line 1
----- tc.v6ingress.bpf.o -----
/tmp/same_tc.v6ingress.o /tmp/committed_tc.v6ingress.o differ: byte 41, line 1
----- tc.v6egress.bpf.o -----
/tmp/same_tc.v6egress.o /tmp/committed_tc.v6egress.o differ: byte 41, line 1
----- v4events.bpf.o -----
/tmp/same_v4events.o /tmp/committed_v4events.o differ: byte 41, line 1
----- v6events.bpf.o -----
/tmp/same_v6events.o /tmp/committed_v6events.o differ: byte 41, line 1
```

parlakisik:
```
CROSS-NAMESPACE COLLISION — PROGRAM ID EVIDENCE
Legacy image: v1.3.4-eksbuild.1   Fix image: latest-change
Node: ip-10-10-7-223   Both pods (ns1/deny-me-0, me-ns1/deny-0) co-scheduled

=== PINNED PROGRAM FAMILIES ===
PRE  : only deny-me-ns1_*  (ONE family shared by TWO pods)
POST : deny-me@ns1_*  AND  deny@me-ns1_*  (TWO families, one per pod)

=== TC PROGRAM IDs PER POD VETH ===
interface        pod                    PRE ing/egr        POST ing/egr      
eni22b168d2b96   ns1/deny-me-0 (victim) 365 / 363          461 / 460         
eni82da2737e58   me-ns1/deny-0 (attacker) 365 / 363          458 / 456         
enia57dbd56ff2   probe/probe            376 / 364          459 / 457         

COLLISION: PRE victim+attacker both = prog 365/363 (SAME program).
FIXED    : POST victim=461/460, attacker=458/456 (DISTINCT programs).

=== RAW PRE SNAPSHOT ===
--- programs
deny-me-ns1_handle_egress
deny-me-ns1_handle_ingress
probe-probe_handle_egress
probe-probe_handle_ingress
--- maps
deny-me-ns1_cp_egress_map
deny-me-ns1_cp_ingress_map
deny-me-ns1_egress_map
deny-me-ns1_egress_pod_state_map
deny-me-ns1_ingress_map
deny-me-ns1_ingress_pod_state_map
global_aws_conntrack_map
global_policy_events
probe-probe_cp_egress_map
probe-probe_cp_ingress_map
probe-probe_egress_map
probe-probe_egress_pod_state_map
probe-probe_ingress_map
probe-probe_ingress_pod_state_map
--- marker (host /var/run/aws-node/.npa_format_v2)
(no marker)
--- tc filters (prog IDs attached to interfaces)
  eni22b168d2b96: tc_ingress_id=365 tc_egress_id=363
  eni82da2737e58: tc_ingress_id=365 tc_egress_id=363
  enia57dbd56ff2: tc_ingress_id=376 tc_egress_id=364

=== RAW POST SNAPSHOT ===
--- programs
deny-me@ns1_handle_egress
deny-me@ns1_handle_ingress
deny@me-ns1_handle_egress
deny@me-ns1_handle_ingress
probe@probe_handle_egress
probe@probe_handle_ingress
--- maps
deny-me@ns1_cp_egress_map
deny-me@ns1_cp_ingress_map
deny-me@ns1_egress_map
deny-me@ns1_egress_pod_state_map
deny-me@ns1_ingress_map
deny-me@ns1_ingress_pod_state_map
deny@me-ns1_cp_egress_map
deny@me-ns1_cp_ingress_map
deny@me-ns1_egress_map
deny@me-ns1_egress_pod_state_map
deny@me-ns1_ingress_map
deny@me-ns1_ingress_pod_state_map
global_aws_conntrack_map
global_policy_events
probe@probe_cp_egress_map
probe@probe_cp_ingress_map
probe@probe_egress_map
probe@probe_egress_pod_state_map
probe@probe_ingress_map
probe@probe_ingress_pod_state_map
--- marker (host /var/run/aws-node/.npa_format_v2)
-rw-r--r--. 1 root root 3 Jun  9 21:30 /host-runaws/.npa_format_v2
--- tc filters (prog IDs attached to interfaces)
  eni22b168d2b96: tc_ingress_id=461 tc_egress_id=460
  eni82da2737e58: tc_ingress_id=458 tc_egress_id=456
  enia57dbd56ff2: tc_ingress_id=459 tc_egress_id=457

```

pkg/ebpf/c/v4events.bpf.c
```diff
@@ -56,4 +56,5 @@ struct bpf_map_def_pvt SEC("maps") policy_events = {
     .pinning = PIN_GLOBAL_NS,
 };
 
+const volatile __u32 NPA_FILE_VERSION = 2;
```
Pavani-Panakanti: we don't need the change in events files. Changing events binary will not recover global conntrack maps. As we just need bpf prog updates we can make changes only in tc files and let conntrack files be recovered

Pavani-Panakanti (approved):

PR #598, opened 2026-06-22
fix: restore egress catch-all in cleanupPod for Ingress-only pods (#591)

### LINKED ISSUE #591: Egress silently dropped after deleting an Ingress-only NetworkPolicy — egress_map missing 0.0.0.0/0 catch-all (vpc-cni v1.21.1-eksbuild.5 / agent v1.3.1-eksbuild.1)
**What happened**:

When a pod is selected only by Ingress-only `NetworkPolicy` resources (no Egress policy applies), egress should be unrestricted. After **a single Argo CD prune** of one such Ingress-only policy (with `prune propagation = foreground`), the per-pod `egress_map` of every pod in that namespace ends up with the `0.0.0.0/0` LPM catch-all entry **missing**, while the pod's `egress_pod_state_map` still indicates DEFAULT_ALLOW for both tiers (Key 0 → 0, Key 1 → 1).

Net effect: the only LPM entry left in `egress_map` is the node IP `/32`. The eBPF egress program LPM-looks up the destination, misses, and drops. **All egress** from selected pods is silently dropped (DNS, intra-NS ClusterIP / PodIP, inter-NS, external). Ingress to those same pods continues to work.

This looks like a different manifestation of the class of bugs in #570 and #573: agent userspace reaches a state where the per-pod state map says DEFAULT_ALLOW but the per-pod allow-list lacks the catch-all entry that DEFAULT_ALLOW requires, and the eBPF program faithfully enforces the inconsistency.

> **Update — Argo syncOption mitigations confirmed insufficient**
>
> First repro hypothesis: the bug is caused by Argo's SSA-driven cascade re-apply of sibling NPs during a prune wave. We tested `ApplyOutOfSyncOnly=true` + `PruneLast=true` to suppress that cascade.
>
> The flags worked as intended — Argo's `syncResult` shows only the single pruned resource, and the `creationTimestamp` of every other NP in the namespace is unchanged — but **the bug still reproduces with identical symptoms**. So the root trigger is not the cascade re-apply; it is the **prune itself**, on its own. Even a clean prune of one Ingress-only NetworkPolicy with all other NPs left untouched is sufficient to evict the `0.0.0.0/0` catch-all from `egress_map` of pods in that namespace.

**Side-by-side dataplane evidence** (same node, same NP set in both NSes, same eBPF program tag `32456ed5a7c029de` attached to both veths):

Healthy pod (`netpol-test-2/app1`) — Argo did not touch this NS in the breaking sync:

```
$ bpftool map dump pinned /sys/fs/bpf/globals/aws/maps/app1-*-netpol-test-2_egress_map
key:
20 00 00 00 0a e0 20 0d                          # /32, 10.224.32.13 (node IP)
value: fe ...
key:
00 00 00 00 00 00 00 00                          # /0, 0.0.0.0  (catch-all, DEFAULT_ALLOW)
value: fe ...
Found 2 elements
```

Broken pod (`netpol-test-1/app1`) — namespace had one NP pruned by Argo:

```
$ bpftool map dump pinned /sys/fs/bpf/globals/aws/maps/app1-*-netpol-test-1_egress_map
key:
20 00 00 00 0a e0 20 0d                          # /32, 10.224.32.13 (node IP)
value: fe ...
Found 1 element                                  # <-- 0.0.0.0/0 catch-all missing
```

Reproduced on two consecutive prune attempts:
- **First attempt** (`syncOptions: [ServerSideApply=true]`, defaults): broken `egress_map`, all egress dropped from `netpol-test-1` pods.
- **Second attempt** (`syncOptions: [ServerSideApply=true, ApplyOutOfSyncOnly=true, PruneLast=true]`): syncResult shows only the single pruned NP touched; `creationTimestamp` of all 5 sibling NPs unchanged; **same broken `egress_map`, same egress drops**.

`egress_pod_state_map` is identical on both pods (Key 0 → 0, Key 1 → 1).

Live in-pod confirmation — from `netpol-test-1/app1` while broken, every TCP destination (intra-NS pod IP, intra-NS ClusterIP, inter-NS pod IP, `1.1.1.1:443`) and DNS lookups all time out (HTTP code `000`); meanwhile from `netpol-test-2/app1` to the exact same `netpol-test-1/app1` pod IP the connection succeeds (HTTP `200`). So the drop is on the egress side of the affected pod's veth, not on the destination's ingress side.

**Pre-emptive answers to the triage questions usually asked here**

- **Which NetworkPolicy is showing this behavior?** All Ingress-only NPs in `netpol-test-1` are equivalent for the trigger. We tested pruning `allow-cert-manager` twice (under different syncOption combinations) and observed identical breakage each time. The namespace contains six NPs: `default-deny-ingress`, `allow-intra-ns`, `allow-from-netpol-test-2`, `allow-cert-manager`, `allow-dns`, `allow-ingress-controller`. Each NP has `policyTypes: [Ingress]` only, `podSelector: {}` so all pods in the NS are selected.
- **Which TargetPod?** Both pods in `netpol-test-1` were affected:
  - `netpol-test-1/app1-79c4bcd68c-5vnfg` (IP `100.65.0.60`)
  - `netpol-test-1/app2-ddf6f6645-wzfss` (IP `100.65.0.174`)
  - On nodes `ip-10-224-32-13` and `ip-10-224-32-28` respectively. Both broke; both healed after the next reconcile that re-creates the pruned NP.
- **PolicyEndpoints fanout?** 1 PolicyEndpoint per NetworkPolicy in this namespace. Cluster-wide there are 135 NPs / 135 PEs across 86 namespaces / 8 nodes / 219 pods, so this isn't a high-fanout-induced bug — it reproduces with a 1:1 NP:PE ratio.
- **Is there churn in the cluster?** Yes — and the churn pattern matters. Static `kubectl delete networkpolicy <one>` does **not** reproduce, even on the exact same NPs we tested via Argo (we ran it for every NP in `netpol-test-1`, all 12 traffic tests stayed green). The trigger is specifically **Argo's prune via foreground propagation**. The Application:
  ```yaml
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
    syncOptions:
      - ServerSideApply=true
      - ApplyOutOfSyncOnly=true   # second attempt
      - PruneLast=true             # second attempt
  ```
  Sync UI: prune propagation `foreground`, no other flags. Single git commit that comments out one NP from the kustomize `resources:` list.
- **Cluster scale**: 8 nodes / 86 namespaces / 219 pods / 135 NPs / 135 PEs. EKS 1.33 (`eks.37`).

**Reproduction steps**

Two namespaces, two app pods each, only Ingress NetworkPolicies — no Egress policy anywhere in the cluster touches these pods.

1. Verify all 12 traffic combinations pass (intra-NS and inter-NS, both directions): ✅ baseline.
2. Edit the kustomization for `netpol-test-1` to comment out **one** policy (e.g. `allow-cert-manager`), commit, push.
3. Trigger Argo sync (default options: SERVER-SIDE APPLY on, prune propagation `foreground`, no other flags).
4. Re-run the traffic test.

Observed: every test originating from a `netpol-test-1` pod fails with timeout (HTTP code `000`). Every other test passes. `netpol-test-2 → netpol-test-1` ingress works (so `netpol-test-1` pods are reachable; they just can't talk back).

Direct comparison runs:
- `kubectl delete networkpolicy <each>` in turn — all 12/12 stay green.
- Same kustomize change driven by Argo sync — 6/12 fail, exactly all rows where the source pod is in `netpol-test-1`.
- Same again with `ApplyOutOfSyncOnly=true + PruneLast=true` — 6/12 fail, syncResult confirms only the pruned NP was touched.

**Recovery without pod or agent restart**: any further NP write that lands on the namespace (we tested re-adding `allow-cert-manager`, but any non-trivial change works) triggers a new reconcile; after that sync `egress_map` is re-populated with both LPM entries and traffic flows. We have not had to bounce a pod or the agent.

**Attach logs**

The aws-eks-nodeagent container exposes only this on the affected node, repeatedly:

```
Logger.check error: failed to get caller
```

Will email the full `aws-cni-support.sh` log archive plus the cluster ARN to `k8s-awscni-triage@amazon.com` and reference this issue number.

Pruning an Ingress-only NetworkPolicy via Argo CD should not change the egress allow-list of selected pods. The `0.0.0.0/0` catch-all entry in `egress_map` should remain in place (or be re-written) for any pod whose `egress_pod_state_map` indicates DEFAULT_ALLOW. Equivalently, no NP delete event that doesn't introduce an Egress policy should ever be capable of dropping the catch-all entry.

Alternative defense-in-depth: the eBPF egress program could fail-open when the LPM lookup misses *and* `pod_state` says DEFAULT_ALLOW, analogous to Option A in #570.

**Anything else we need to know?**:

- The namespace that was *not* touched by the breaking sync (`netpol-test-2`) keeps a correct `egress_map` and is unaffected. So this isn't a cluster-global state issue; it's per-namespace and triggered by the Argo prune of a NetworkPolicy in that namespace.
- The eBPF program tag is identical on broken and healthy pods (`32456ed5a7c029de`), so the inconsistency lives in map state, not program state.
- We tested two Argo syncOption mitigations: `ApplyOutOfSyncOnly=true` and `PruneLast=true`. Neither prevents the bug. The only remaining Argo-side knob is changing prune propagation from `foreground` → `background` / `orphan`. Will report back.
- Asks:
  1. Is the `0.0.0.0/0` LPM entry on `egress_map` the documented mechanism for expressing DEFAULT_ALLOW egress on a pod that has only Ingress policies? Reading [`pkg/ebpf/c/tc.v4egress.bpf.c`](https://github.com/aws/aws-network-policy-agent/blob/main/pkg/ebpf/c/tc.v4egress.bpf.c) this seems to be the case — but pointers welcome.
  2. Where in the agent does the catch-all entry get (re-)written, and is there a code path on `NetworkPolicy` delete (driven via the `networking.k8s.aws/resources` finalizer) that re-derives `egress_map` for a pod and forgets to add it back?
  3. Anything in v1.4.x / current `main` that already addresses this?

Happy to provide:
- All `bpftool map dump` snapshots (before, broken, recovered).
- Full agent and CNI log archives via the support email.
- The exact NP YAMLs and the Argo Application/AppSet specs.

**Environment**:
- Kubernetes version (use `kubectl version`):
  - Client Version: v1.35.3
  - Server Version: v1.33.11-eks-40737a8 (EKS platform `eks.37`)
- CNI Version: `amazon-k8s-cni:v1.21.1-eksbuild.5` (EKS addon `vpc-cni v1.21.1-eksbuild.5`)
- Network Policy Agent Version: `aws-network-policy-agent:v1.3.1-eksbuild.1`
- Addon configurationValues:
  ```
  enableNetworkPolicy: "true"
  env:
    AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG: "true"
    ENABLE_POD_ENI: "false"
    ENABLE_PREFIX_DELEGATION: "true"
    ENABLE_SUBNET_DISCOVERY: "true"
    ENI_CONFIG_LABEL_DEF: topology.kubernetes.io/zone
  ```
  Agent env: `NETWORK_POLICY_ENFORCING_MODE: standard`, `--enable-network-policy=true`.
- OS: Bottlerocket OS 1.59.0 (aws-k8s-1.33).
- Kernel: `6.12.79`. Container runtime: `containerd 2.1.6+bottlerocket`.

RBROE00094:
## What

Restores the `0.0.0.0/0` egress catch-all entry in `cleanupPod` for pods that are still selected by other NetworkPolicies but have no active Egress policy.

## Why

The apply path (`reconcilePolicyEndpoint`, [L218–L227](https://github.com/aws/aws-network-policy-agent/blob/main/controllers/policyendpoints_controller.go#L218-L227)) is symmetric — it adds a catch-all to whichever rule set is empty *and* not isolated:

```go
if len(ingressRules) == 0 && !isIngressIsolated {
    r.addCatchAllEntry(&ingressRules)
}
if len(egressRules) == 0 && !isEgressIsolated {
    r.addCatchAllEntry(&egressRules)
}
```

The cleanup path (`cleanupPod`, [L391–L401](https://github.com/aws/aws-network-policy-agent/blob/main/controllers/policyendpoints_controller.go#L391-L401)) is asymmetric — it adds the catch-all to `ingressRules` only:

```go
} else {
    log().Infof("Active policies against this pod. Skip Detaching probes and Update Maps... ")
    if noActiveIngressPolicies {
        r.addCatchAllEntry(&ingressRules)
    }
    // missing: symmetric branch for noActiveEgressPolicies
    err = r.updateeBPFMaps(podIdentifier, ingressRules, egressRules, ebpf.POLICIES_APPLIED)
}
```

For a pod selected by N Ingress-only NetworkPolicies (`policyTypes: [Ingress]` only — the default and overwhelmingly common case), after deleting one of them:

- `deriveIngressAndEgressFirewallRules` returns `ingressRules=[N-1 rules]`, `egressRules=[]`, `isEgressIsolated=false`
- `noActiveIngressPolicies=false`, `noActiveEgressPolicies=true`
- Else branch entered; ingress catch-all not added (no-op since ingressRules is non-empty); **egress catch-all never written**
- `updateeBPFMaps(..., POLICIES_APPLIED)` writes the empty egress rule set → `egress_map` loses `0.0.0.0/0`
- `egress_pod_state_map` correctly stays at `POLICIES_APPLIED`

The eBPF program in [`pkg/ebpf/c/tc.v4egress.bpf.c::evaluateNamespacePolicyByLookUp`](https://github.com/aws/aws-network-policy-agent/blob/main/pkg/ebpf/c/tc.v4egress.bpf.c#L154-L165) then drops everything:

```c
struct lpm_trie_val *trie_val = bpf_map_lookup_elem(&egress_map, &trie_key);
if (trie_val == NULL) {
    if (pod_state != POLICIES_APPLIED) {
        return ACTION_PASS;
    }
    return ACTION_DENY;          // <-- hit: pod_state == POLICIES_APPLIED, egress_map empty
}
```

Net effect: silent egress black-hole on every pod the deleted Ingress NP selected, until the next NetworkPolicy write to that namespace triggers `reconcilePolicyEndpoint` (apply path, symmetric) which re-writes the catch-all.

GitOps controllers with `selfHeal: true` mask this in production because their re-apply loop heals the breakage within 1–3 seconds — long enough to drop in-flight egress connections (visible as transient DNS/connect failures correlated with NP churn), short enough to escape HTTP-level synthetic tests. Triggering the bug via `kubectl delete` on a cluster with selfHeal disabled produces persistent, deterministic egress drop until any subsequent NP write.

## Fix

Add the missing symmetric branch in `cleanupPod`:

```go
if noActiveEgressPolicies {
    log().Info("No Egress rules and no egress isolation - Appending catch all entry")
    r.addCatchAllEntry(&egressRules)
}
```

This mirrors the existing logic in `reconcilePolicyEndpoint` and re-establishes the invariant that `egress_map` and `egress_pod_state_map` agree on the pod's egress mode.

## Tests

New unit test `TestCleanupPod_PreservesCatchAllOnRemainingPolicy` exercises the exact bug:
- Pod selected by two Ingress-only NPs
- One is cleaned up; the other remains
- Asserts the egress rules written to the bpf client contain `0.0.0.0/0`
- Asserts ingress rules from the remaining NP are written unchanged

Verified the test **fails on `main`** and **passes with the fix**.

Also added `LastIngressRules` / `LastEgressRules` capture fields on `MockBpfClient` so future tests can assert against rule-map writes. Existing tests (`TestAddCatchAllEntry`, `TestPolicyEndpointReconcile`, `TestDeriveIngressAndEgressFirewallRules`, `TestUpdateeBPFMaps_FailClosed`, `TestUpdateClusterPolicyBPFMaps_FailClosed`, etc.) continue to pass — the capture fields default to nil and don't affect existing call-log assertions.

## Related

Closes #591. Verified in EKS 1.33 with agent `v1.3.1-eksbuild.1` and the workaround documented in the issue (explicit `allow-egress-all` NP) is no longer needed after this fix.

cc @jaydeokar — per your comment on the issue.

### DIFF
--- a/controllers/policyendpoints_controller.go
+++ b/controllers/policyendpoints_controller.go
@@ -394,10 +394,18 @@ func (r *PolicyEndpointsReconciler) cleanupPod(ctx context.Context, targetPod np
 			// Update the Maps and move on
 			log().Infof("Active policies against this pod. Skip Detaching probes and Update Maps... ")
 			if noActiveIngressPolicies {
-				// No active ingress rules for this pod, but we only should land here
-				// if there are active egress rules. So, we need to add an allow-all entry to ingress rule set
+				// No active ingress rules for this pod (only egress rules remain).
+				// Add an allow-all entry to ingress rule set so non-egress traffic isn't dropped.
 				log().Info("No Ingress rules and no ingress isolation - Appending catch all entry")
 				r.addCatchAllEntry(&ingressRules)
+			}
+			if noActiveEgressPolicies {
+				// No active egress rules for this pod (only ingress rules remain).
+				// Add an allow-all entry to egress rule set so egress isn't dropped.
+				// Without this, egress_map ends up empty while egress_pod_state_map stays
+				// at POLICIES_APPLIED, and the eBPF dataplane drops all egress traffic.
+				log().Info("No Egress rules and no egress isolation - Appending catch all entry")
+				r.addCatchAllEntry(&egressRules)
 			}
 
 			err = r.updateeBPFMaps(podIdentifier, ingressRules, egressRules, ebpf.POLICIES_APPLIED)
--- a/controllers/policyendpoints_controller.go
+++ b/controllers/policyendpoints_controller.go
@@ -394,16 +394,16 @@ func (r *PolicyEndpointsReconciler) cleanupPod(ctx context.Context, targetPod np
 			// Update the Maps and move on
 			log().Infof("Active policies against this pod. Skip Detaching probes and Update Maps... ")
 			if noActiveIngressPolicies {
-				// No active ingress rules for this pod (only egress rules remain).
-				// Add an allow-all entry to ingress rule set so non-egress traffic isn't dropped.
+				// Add an allow-all entry to ingress rule set so ingress traffic
+				// isn't dropped by the eBPF dataplane.
 				log().Info("No Ingress rules and no ingress isolation - Appending catch all entry")
 				r.addCatchAllEntry(&ingressRules)
 			}
 			if noActiveEgressPolicies {
-				// No active egress rules for this pod (only ingress rules remain).
-				// Add an allow-all entry to egress rule set so egress isn't dropped.
-				// Without this, egress_map ends up empty while egress_pod_state_map stays
-				// at POLICIES_APPLIED, and the eBPF dataplane drops all egress traffic.
+				// Add an allow-all entry to egress rule set so egress traffic
+				// isn't dropped by the eBPF dataplane. Without this, egress_map
+				// ends up not having the catch all rule while pod_state_map stays
+				// at POLICIES_APPLIED and the eBPF dataplane drops all egress traffic.
 				log().Info("No Egress rules and no egress isolation - Appending catch all entry")
 				r.addCatchAllEntry(&egressRules)
 			}
--- a/controllers/policyendpoints_controller_test.go
+++ b/controllers/policyendpoints_controller_test.go
@@ -1437,3 +1437,121 @@ func TestCleanupPod_PreservesCatchAllOnRemainingPolicy(t *testing.T) {
 	assert.NotContains(t, mockBpf.LastIngressRules, catchAll)
 	assert.NotEmpty(t, mockBpf.LastIngressRules)
 }
+
+// Symmetric variant of TestCleanupPod_PreservesCatchAllOnRemainingPolicy:
+// after cleaning up one of N Egress-only NPs selecting a pod, the ingress rules
+// written to the bpf client must contain the 0.0.0.0/0 catch-all so that
+// ingress_map stays consistent with pod_state=POLICIES_APPLIED.
+func TestCleanupPod_EgressOnlyAddsIngressCatchAll(t *testing.T) {
+	protocolTCP := corev1.ProtocolTCP
+	var port80 int32 = 80
+
+	namespace := "my-namespace"
+	podIdentifier := "deployment1rs@my-namespace"
+	cleanedUpPE := "egress-1-abcd"
+	remainingPE := "egress-2-abcd"
+
+	remainingPolicy := policyendpoint.PolicyEndpoint{
+		ObjectMeta: metav1.ObjectMeta{Name: remainingPE, Namespace: namespace},
+		Spec: policyendpoint.PolicyEndpointSpec{
+			PodSelector: &metav1.LabelSelector{},
+			PolicyRef:   policyendpoint.PolicyReference{Name: "egress-2", Namespace: namespace},
+			Egress: []policyendpoint.EndpointInfo{
+				{
+					CIDR:  "10.0.0.0/24",
+					Ports: []policyendpoint.Port{{Port: &port80, Protocol: &protocolTCP}},
+				},
+			},
+		},
+	}
+
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockClient := mock_client.NewMockClient(ctrl)
+	mockBpf := &ebpf.MockBpfClient{}
+	r := NewPolicyEndpointsReconciler(mockClient, "1.1.1.1", mockBpf, false)
+	r.podIdentifierToPolicyEndpointMap.Store(podIdentifier, []string{remainingPE})
+
+	mockClient.EXPECT().Get(gomock.Any(), types.NamespacedName{Name: remainingPE, Namespace: namespace}, gomock.Any()).DoAndReturn(
+		func(ctx context.Context, key types.NamespacedName, currentPE *policyendpoint.PolicyEndpoint, opts ...client.GetOption) error {
+			*currentPE = remainingPolicy
+			return nil
+		},
+	).AnyTimes()
+
+	targetPod := npatypes.Pod{
+		NamespacedName: types.NamespacedName{Name: "deployment1rs-1", Namespace: namespace},
+		PodIP:          "10.1.1.1",
+	}
+
+	err := r.cleanupPod(context.Background(), targetPod, cleanedUpPE, true)
+	assert.NoError(t, err)
+
+	catchAll := fwrp.EbpfFirewallRules{IPCidr: "0.0.0.0/0"}
+	assert.Contains(t, mockBpf.LastIngressRules, catchAll)
+	assert.NotContains(t, mockBpf.LastEgressRules, catchAll)
+	assert.NotEmpty(t, mockBpf.LastEgressRules)
+}
+
+// When the remaining PolicyEndpoint has BOTH ingress and egress rules,
+// neither side should get a 0.0.0.0/0 catch-all appended on cleanup —
+// the existing rule sets are non-empty so the dataplane will not drop traffic.
+func TestCleanupPod_BothDirectionsActiveNoCatchAll(t *testing.T) {
+	protocolTCP := corev1.ProtocolTCP
+	var port80 int32 = 80
+
+	namespace := "my-namespace"
+	podIdentifier := "deployment1rs@my-namespace"
+	cleanedUpPE := "ingress-1-abcd"
+	remainingPE := "mixed-2-abcd"
+
+	remainingPolicy := policyendpoint.PolicyEndpoint{
+		ObjectMeta: metav1.ObjectMeta{Name: remainingPE, Namespace: namespace},
+		Spec: policyendpoint.PolicyEndpointSpec{
+			PodSelector: &metav1.LabelSelector{},
+			PolicyRef:   policyendpoint.PolicyReference{Name: "mixed-2", Namespace: namespace},
+			Ingress: []policyendpoint.EndpointInfo{
+				{
+					CIDR:  "10.0.0.0/24",
+					Ports: []policyendpoint.Port{{Port: &port80, Protocol: &protocolTCP}},
+				},
+			},
+			Egress: []policyendpoint.EndpointInfo{
+				{
+					CIDR:  "10.0.1.0/24",
+					Ports: []policyendpoint.Port{{Port: &port80, Protocol: &protocolTCP}},
+				},
+			},
+		},
+	}
+
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockClient := mock_client.NewMockClient(ctrl)
+	mockBpf := &ebpf.MockBpfClient{}
+	r := NewPolicyEndpointsReconciler(mockClient, "1.1.1.1", mockBpf, false)
+	r.podIdentifierToPolicyEndpointMap.Store(podIdentifier, []string{remainingPE})
+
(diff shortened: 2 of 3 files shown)


controllers/policyendpoints_controller.go
```diff
@@ -394,11 +394,19 @@ func (r *PolicyEndpointsReconciler) cleanupPod(ctx context.Context, targetPod np
...
-				// if there are active egress rules. So, we need to add an allow-all entry to ingress rule set
+				// No active ingress rules for this pod (only egress rules remain).
+				// Add an allow-all entry to ingress rule set so non-egress traffic isn't dropped.
 				log().Info("No Ingress rules and no ingress isolation - Appending catch all entry")
 				r.addCatchAllEntry(&ingressRules)
 			}
+			if noActiveEgressPolicies {
+				// No active egress rules for this pod (only ingress rules remain).
+				// Add an allow-all entry to egress rule set so egress isn't dropped.
+				// Without this, egress_map ends up empty while egress_pod_state_map stays
```
jaydeokar: nit - Without this, egress_map ends up not having the catch all rule while pod_state_map stays at POLICIES_APPLIED and the eBPF dataplane drops all egress traffic

controllers/policyendpoints_controller.go
```diff
@@ -394,11 +394,19 @@ func (r *PolicyEndpointsReconciler) cleanupPod(ctx context.Context, targetPod np
 			// Update the Maps and move on
 			log().Infof("Active policies against this pod. Skip Detaching probes and Update Maps... ")
 			if noActiveIngressPolicies {
-				// No active ingress rules for this pod, but we only should land here
-				// if there are active egress rules. So, we need to add an allow-all entry to ingress rule set
+				// No active ingress rules for this pod (only egress rules remain).
+				// Add an allow-all entry to ingress rule set so non-egress traffic isn't dropped.
```
jaydeokar: Let's be explicit here and say "Add an allow-all entry to ingress rule set so ingress traffic isn't dropped by the eBPF dataplane"

jaydeokar:
Added nits for comments - Rest looks good to me

controllers/policyendpoints_controller_test.go
```diff
@@ -1381,3 +1381,59 @@ func TestUpdateClusterPolicyBPFMaps_FailClosed(t *testing.T) {
 		return r.updateClusterPolicyBPFMaps("test-pod-id", nil, nil)
 	})
 }
+
+// Regression test for the asymmetric catch-all bug in cleanupPod (issue #591):
+// after cleaning up one of N Ingress-only NPs selecting a pod, the egress rules
+// written to the bpf client must still contain the 0.0.0.0/0 catch-all so that
+// egress_map stays consistent with pod_state=POLICIES_APPLIED.
+func TestCleanupPod_PreservesCatchAllOnRemainingPolicy(t *testing.T) {
```
Pavani-Panakanti, line 1389: Thanks for these changes. Can you expand tests to cover other two similar variants as well - egress-only --> ingress catch all and both directions active --> no catch all

jaydeokar:
lgtm

jaydeokar (approved):

PR #604, opened 2026-07-06
feat: embed version and build metadata in NPA binary

nakshah87:
## Summary
Stamp version and build metadata into the NPA binary via ldflags so it is logged at startup for runtime diagnosis.
  
Changes:
  - pkg/version/version.go: New file with GitVersion, BuildDate, EbpfSdkVersion variables (defaulting to "unknown") and a String() helper
  - main.go: Startup log now includes version.String() output
  - Makefile: Added VERSION_LDFLAGS that injects git version, build date, and ebpf SDK version at compile time

At build time, the Makefile resolves:
  | Variable | Source |
  |----------|--------|
  | `GIT_VERSION` | `git describe --tags --always --dirty` |
  | `BUILD_DATE` | `date -u +%Y-%m-%dT%H:%M:%SZ` |
  | `EBPF_SDK_VERSION` | `go list -m -f '{{.Version}}' github.com/aws/aws-ebpf-sdk-go` |
  
  These are injected into the binary via ldflags (`-X pkg/version.GitVersion=...`).
  
  If any command fails (e.g., no .git directory), the value defaults to "unknown".

  ## Testing 
  - Built locally with Go 1.26.4, confirmed startup log shows all metadata fields
  - Deployed test image to dev cluster, verified log output in /var/log/aws-routed-eni/network-policy-agent.log
  - Unstamped binary (no ldflags) correctly shows unknown (built: unknown, ebpf-sdk: unknown)
 

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/main.go
+++ b/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/aws/aws-network-policy-agent/pkg/rpcclient"
 	"github.com/aws/aws-network-policy-agent/pkg/utils"
 	"github.com/aws/aws-network-policy-agent/pkg/utils/imds"
+	"github.com/aws/aws-network-policy-agent/pkg/version"
 	"github.com/samber/lo"
 	"google.golang.org/protobuf/types/known/emptypb"
 
@@ -74,7 +75,7 @@ func main() {
 		os.Exit(1)
 	}
 	log := logger.New(ctrlConfig.LogLevel, ctrlConfig.LogFile, ctrlConfig.LogFileMaxSize, ctrlConfig.LogFileMaxBackups)
-	log.Infof("Starting network policy agent with log level: %s", ctrlConfig.LogLevel)
+	log.Infof("Starting network policy agent: %s, log level: %s", version.String(), ctrlConfig.LogLevel)
 
 	ctrl.SetLogger(logger.GetControllerRuntimeLogger())
 	if ctrlConfig.EnableProfiling {
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,7 +1,13 @@
 package version
 
+import "fmt"
+
 var (
-	GitVersion string
-	GitCommit  string
-	BuildDate  string
+	GitVersion     = "unknown"
+	BuildDate      = "unknown"
+	EbpfSDKVersion = "unknown"
 )
+
+func String() string {
+	return fmt.Sprintf("%s (built: %s, ebpf-sdk: %s)", GitVersion, BuildDate, EbpfSDKVersion)
+}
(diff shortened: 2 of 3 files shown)


pkg/version/version.go
```diff
@@ -1,7 +1,13 @@
 package version
 
+import "fmt"
+
 var (
-	GitVersion string
-	GitCommit  string
```
jupdec, line 5: why do we need to remove GitCommit?

nakshah87: We can keep it but I did not see any additional helpful info we could get out of displaying commit hash.
Also not a breaking change since these were never used.

main.go
```diff
@@ -178,20 +180,28 @@ func loadControllerConfig() (config.ControllerConfig, error) {
 	fs := pflag.NewFlagSet("", pflag.ExitOnError)
 	controllerConfig.BindFlags(fs)
 
+	showVersion := fs.Bool("version", false, "Print version information and exit")
```
jaydeokar: "Print version and exit".. Don't think we can do this while the controller is running as a container.

nakshah87: Good point — you're right that --version with exit won't be used during normal container operation. The use case is image verification outside the running controller:

The version is logged right at the beginning. If the logs roll over we can loose that detail. 

We can run the below cli to get the version details.
$   kubectl run npa-version-check \
    --image=<image> \
    --restart=Never --rm -i --command -- /controller --version

nakshah87: Also,
* this displays the ebpf sdk version
* can be extended in future to display any additional versions.

main.go
```diff
@@ -178,10 +181,17 @@ func loadControllerConfig() (config.ControllerConfig, error) {
...
+
 	if err := fs.Parse(os.Args); err != nil {
 		return controllerConfig, err
 	}
 
+	if *showVersion {
+		fmt.Println(version.String())
```
viveksb007: why are we doing `fmt.Println` and not using existing log instance

main.go
```diff
@@ -75,7 +77,8 @@ func main() {
 		os.Exit(1)
 	}
 	log := logger.New(ctrlConfig.LogLevel, ctrlConfig.LogFile, ctrlConfig.LogFileMaxSize, ctrlConfig.LogFileMaxBackups)
-	log.Infof("Starting network policy agent with log level: %s", ctrlConfig.LogLevel)
+	log.Infof("Starting network policy agent: %s", version.String())
```
viveksb007: nit - can club log level info in same line

main.go
viveksb007: where is this "version" flag coming from?

I don't think we even need to print version here in this code block, this code part seems rock solid, it something fails here, it will be caught in early testing phase

main.go
nakshah87: --version is handled in loadControllerConfig() which runs before the logger is initialized — there's no log instance available at that point.
  
  Beyond that, for --version we want to print plain text to stdout and exit.
  Using the logger would write JSON to /var/log/aws-routed-eni/ instead of the terminal, and would fail if that directory doesn't exist (e.g., in a standalone kubectl run pod used for image verification).

viveksb007 (approved):

viveksb007 (approved):
lgtm

PR #605, opened 2026-07-06
Fix stale eBPF entries when namespace label is removed

jupdec:
**Issue #, if available:*
 When a namespace label matching a ClusterNetworkPolicy's subject selector is removed, the nodeagent must clear the eBPF deny entries for pods that are no longer targeted. Previously the entries survived until the pod restarted, leaving ingress denied indefinitely even though nothing selected the pod.

  *Description of changes:*

`deriveTargetPodsForParentCNP` mutated `podIdentifierToClusterPolicyEndpointMap` and `clusterNetworkPolicyToPodIdentifierMap` inline before `cleanupClusterPolicyPod` ran. `cleanupClusterPolicyPod` used the first map as a gate — if the entry was absent, it fell through to `return nil` without touching eBPF. So the "no CPE targets this identifier anymore" case (exactly the one that needs cleanup) silently no-oped.

   This PR:

  1. **`deriveTargetPodsForParentCNP`** now also returns `parentCPEList` and only mutates
  `podIdentifierToClusterPolicyEndpointMap` — scrubbing the parent's CPEs from every stale identifier's entry so the
  presence check in `cleanupClusterPolicyPod` only sees sibling CPEs from OTHER parents.
  `ClusterPolicyEndpointSelectorMap` and `clusterNetworkPolicyToPodIdentifierMap` are no longer mutated inline.

  2. **`commitClusterPolicyEndpointState`** (new helper) commits the two remaining map updates AFTER
  `updateClusterPolicyEnforcementStatusForPods` returns success. If cleanup errors, the maps stay pre-cleanup and the requeued reconcile re-derives the same stale set — `return err` now actually retries the clear instead of no-oping.

  3. **`cleanupClusterPolicyPod`** — added an `else` branch that fires when the identifier's entry has been fully scrubbed (truly-stale case). It calls `updateClusterPolicyBPFMaps(podIdentifier, nil, nil)`, which bulk-clears the
  cluster-policy maps and resets pod state to `DEFAULT_ALLOW` (or `DEFAULT_DENY` in strict mode via
  `updateClusterPolicyBPFMaps`' existing logic).

  4. **`HasBPFContext`** (new `BpfClient` interface method) — the else branch skips the clear when the eBPF context is already detached (last pod of the identifier left the node). Without this, routine pod deletion would call
  `UpdateClusterPolicyEbpfMaps` on a missing context and fail the reconcile forever.

  5. Integration test rewrite (`test/integration/clusternetworkpolicy/stale_rule_cleanup_test.go`) — toggles labels on the target **namespace** (matching the CNP's `namespaceSelector` subject) rather than the deployment pod template, so the existing pod is unchanged and its stale eBPF entries are what the test exercises.

## Test steps: 
  
Unit tests (`controllers/clusterpolicyendpoints_controller_test.go`):
  - Stale pod identifier's eBPF maps are cleared and pod state reset to `DEFAULT_ALLOW` when the CPE keeps existing but stops selecting the pod.
  - Stale pod with no eBPF context registered is skipped, not errored — routine pod deletion must not fail the reconcile.
  - Sibling CPE of the same parent CNP does not keep stale rules alive.
  - Active pod identifier's rules remain when there are no stale identifiers.

  Integration test (`test/integration/clusternetworkpolicy/stale_rule_cleanup_test.go`):
  - Two CNPs select the same server pod (one under test, one unrelated control). The subject label of the CNP under test is removed from the namespace; the unrelated CNP's Deny rule must still apply. Then the unrelated label is removed; the pod must become reachable again without restart, while both CNPs and their ClusterPolicyEndpoints keep existing.

    
  **Manual verification** on a fresh EKS cluster (K8s 1.31, us-west-2, VPC CNI v1.23.0, NPC on dataplane):

  | Scenario | Mode | Result |
  |---|---|---|
  | Two CNPs on one pod, labels removed one at a time — subject first, then unrelated | standard | **PASS** — 1 spec
  passed in 62.6s |
  | Two CNPs on one pod, labels removed one at a time — subject first, then unrelated | strict | **PASS** — 1 spec passed in 64.6s |
  | Pod deletion (detached probes, no eBPF context) | standard | **PASS** — `HasBPFContext` guard skips cleanly |

  | Two CNPs on one pod, labels removed one at a time — subject first, then unrelated | standard | **PASS** — 1 spec
  passed in 62.6s |
  | Two CNPs on one pod, labels removed one at a time — subject first, then unrelated | strict | **PASS** — 1 spec passed in 64.6s |
  | Pod deletion (detached probes, no eBPF context) | standard | **PASS** — `HasBPFContext` guard skips cleanly |
  
  `aws-node` pod restart count remained 0 throughout; no reconcile errors observed.

  Nodeagent log excerpts confirming the new code paths fire:
  
  **Truly-stale clear path — standard mode:**
```
  level=debug msg="No cluster policies left for podIdentifier stale-server-f8b8c7777@stale-server, clearing cluster policy maps"
 ```
  **Truly-stale clear path — strict mode (subsequent pod-state → DEFAULT_DENY, value: 1):**

```
  00:36:08.485  debug  No cluster policies left for podIdentifier stale-server-f8b8c7777@stale-server, clearing cluster
  policy maps
  00:36:08.485  info   Pod has an Ingress hook attached. Update the corresponding map progFD: 22, mapName:
  ingress_pod_state_map, key: 1, value: 1
  00:36:08.485  info   Pod has an Egress hook attached.  Update the corresponding map progFD: 26, mapName:
  egress_pod_state_map,  key: 1, value: 1
```
  **Pod deletion — `HasBPFContext` guard skips the clear when probes are already detached:**
```
  level=debug msg="Skipping cluster policy cleanup for podIdentifier pd@pd-test: no eBPF context registered"
```

### CODE REVIEW

**Comment by parlakisik** [comment]
lgtm , lets validate with tests

File: controllers/clusterpolicyendpoints_controller.go [hunk 1]
```diff
@@ -164,7 +168,18 @@ func (r *ClusterPolicyEndpointsReconciler) reconcileClusterPolicyEndpoint(ctx co
...
-	targetPods, targetPodIdentifiers, podsToBeCleanedUp := r.deriveTargetPodsForParentCNP(ctx, parentCNP, resourceName)
+	targetPods, targetPodIdentifiers, podsToBeCleanedUp, stalePodIdentifiers := r.deriveTargetPodsForParentCNP(ctx, parentCNP, resourceName)
+
+	// For stale pod identifiers: remove from lookup map AND clear their eBPF maps.
+	// Must delete from map first so cleanup won't re-derive rules from this CPE,
+	// then explicitly clear eBPF entries since cleanupClusterPolicyPod skips pods
+	// not in the lookup map.
+	for _, podIdentifier := range stalePodIdentifiers {
+		utils.DeletePolicyEndpointFromPodIdentifierMap(&r.podIdentifierToClusterPolicyEndpointMap, &r.podIdentifierToClusterPolicyEndpointMapMutex, podIdentifier, resourceName)
+		if err := r.updateClusterPolicyBPFMaps(podIdentifier, nil, nil); err != nil {
```
jaydeokar: We cannot directly pass ingress and egress as `nil` when we delete stale pod identifiers

An identifier can be selected by multiple network policy (aka policyendpoints), so it might still have rules that needs to be applied

yash97:
Could we test the scenario where a single parent CNP has two CPEs selecting the same pod, then the pod stops matching that parent? Please add an integration test verifying that both CPEs’ rules are removed while rules from unrelated parent CNPs are preserved.

controllers/clusterpolicyendpoints_controller.go
```diff
@@ -453,6 +460,26 @@ func (r *ClusterPolicyEndpointsReconciler) cleanupClusterPolicyPod(ctx context.C
...
+		// pod deletion fail the reconcile and requeue forever.
+		if !r.ebpfClient.HasBPFContext(podIdentifier) {
+			log().Debugf("Skipping cluster policy cleanup for podIdentifier %s: no eBPF context registered", podIdentifier)
+			return nil
+		}
+
+		log().Infof("No cluster policies left for podIdentifier %s, clearing cluster policy maps", podIdentifier)
+		if err := r.updateClusterPolicyBPFMaps(podIdentifier, nil, nil); err != nil {
+			log().Errorf("cluster policy map clear failed for podIdentifier %s: %v", podIdentifier, err)
+			return err
```
Pavani-Panakanti: retry here won't work because we delete the CPE resource here - https://github.com/aws/aws-network-policy-agent/blob/d967c8e9379a84887c4e84ea4115f0c56b474cfd/controllers/clusterpolicyendpoints_controller.go#L392 
When pods got removed first time -> This PE has no pods so remove it from mapping. Also trigger cleanup on the removed pods 
After retry --> this CPE has no pods now and earlier as well (because we cleaned up). Nothing to do. No cleanup of the retry pod is triggered

Pavani-Panakanti: I see deleting those maps based on success or failure here would need bunch of rewrite - do evaluate this and see if we want to take this retry

Pavani-Panakanti: we will also skip the rest of enforcement here https://github.com/aws/aws-network-policy-agent/blob/d967c8e9379a84887c4e84ea4115f0c56b474cfd/controllers/clusterpolicyendpoints_controller.go#L170 for a retry that wont work

test/integration/clusternetworkpolicy/stale_rule_cleanup_test.go
```diff
@@ -0,0 +1,303 @@
...
+					return cpesUnderTest
+				}, 60*time.Second, 5*time.Second).ShouldNot(BeEmpty())
+
+				// When the parent CNP is sliced into several CPEs, every one of them has to be
+				// dropped for the identifier. Log the count so a multi-CPE run is visible.
+				GinkgoWriter.Printf("CNP %s produced %d ClusterPolicyEndpoint(s): %v\n",
+					parentUnderTest, len(cpesUnderTest), cpesUnderTest)
+			})
+
+			By("Removing the label so the server pod stops matching the CNP under test", func() {
```
Pavani-Panakanti: Can you validate below and also update PR description with all integration test suite results. 

---

 This patch will be rejected by the API server with a 422, so the test fails at its central action and never reaches any of the assertions that
  validate the fix.

  NewDefaultDeploymentBuilder().Build() uses the same d.labels map for both Spec.Selector.MatchLabels and Spec.Template.ObjectMeta.Labels
  (test/framework/manifest/deployment.go:88-94, the same d.labels reference assigned twice). The AddLabel calls in BeforeAll therefore put subjectLabelKey
  and unrelatedLabelKey into the selector as well as the template. This patch removes the label from the template only, leaving the selector still
  requiring it.

  Kubernetes rejects that. ValidateDeploymentUpdate → ValidateDeploymentSpec → ValidatePodTemplateSpecForReplicaSet fails with `selector` does not match
  template `labels` when the selector does not match the template labels (upstream pkg/apis/apps/validation/validation.go, ValidateDeploymentSpec calling
  ValidatePodTemplateSpecForReplicaSet(&spec.Template, selector, ...); the error is raised at the !selector.Matches(labels) check in that function).
  Deployment selectors are immutable for apps/v1, so patching the selector to match is not an option either.

controllers/clusterpolicyendpoints_controller_test.go
```diff
@@ -0,0 +1,336 @@
...
+		assert.Nil(t, err, "reconcile should not error when the stale pod has no eBPF context to clear")
+
+		// No eBPF map update should have been attempted for the context-less pod.
+		assert.NotContains(t, mockBpf.CallLog, "UpdateClusterPolicyEbpfMaps",
+			"UpdateClusterPolicyEbpfMaps must not be called when there is no eBPF context")
+	})
+
+	t.Run("sibling CPE of the same CNP does not keep stale rules alive", func(t *testing.T) {
+		// A single CNP sliced into two CPEs. The old code removed only the reconciled CPE
+		// name from the stale identifier's entry, leaving the sibling CPE behind so the entry
```
Pavani-Panakanti: This does not look right, even old code was removing all CPE's and no code change was done there. Also lets remove references like old code as this wont be true and confusing as we make more changes. We can just add what is the expected behavior here

controllers/clusterpolicyendpoints_controller.go
```diff
@@ -392,6 +392,13 @@ func (r *ClusterPolicyEndpointsReconciler) deriveTargetPodsForParentCNP(ctx cont
 			r.ClusterPolicyEndpointSelectorMap.Delete(ClusterPolicyEndpointResource)
 		}
 
+		// Remove every CPE of this parent CNP from the stale identifiers' entries. All of
+		// the parent's CPEs must go, not just resourceName: a single podIdentifier holds the
+		// whole parentCPEList (see deriveClusterPolicyTargetPods), so leaving a sibling CPE
+		// behind would keep the entry alive and let cleanupClusterPolicyPod re-derive this
+		// CNP's rules instead of clearing them. Entries belonging to other CNPs are left
+		// intact, so cleanupClusterPolicyPod can tell "no policies left" (entry absent) from
+		// "another policy still applies" (entry present).
```
Pavani-Panakanti: we are just documenting existing behavior ?

controllers/clusterpolicyendpoints_controller.go
```diff
@@ -453,6 +460,26 @@ func (r *ClusterPolicyEndpointsReconciler) cleanupClusterPolicyPod(ctx context.C
...
+		// Skip pods whose eBPF context is already gone: the probes are detached once the last
+		// pod of a podIdentifier leaves the node, so there is nothing left to clear and
+		// UpdateClusterPolicyEbpfMaps would fail. Returning an error there would make routine
+		// pod deletion fail the reconcile and requeue forever.
+		if !r.ebpfClient.HasBPFContext(podIdentifier) {
+			log().Debugf("Skipping cluster policy cleanup for podIdentifier %s: no eBPF context registered", podIdentifier)
+			return nil
+		}
+
+		log().Infof("No cluster policies left for podIdentifier %s, clearing cluster policy maps", podIdentifier)
```
Pavani-Panakanti: should we move this to debug to reduce log noise ?

test/integration/clusternetworkpolicy/stale_rule_cleanup_test.go
```diff
@@ -0,0 +1,303 @@
...
+		var (
+			serverDeployment *appsv1.Deployment
+			clientDeployment *appsv1.Deployment
+			policies         []*unstructured.Unstructured
+			serverPodIP      string
+			clientPodName    string
+		)
+
+		// waitForServerRolloutAndRefreshIP waits until the server deployment has finished rolling
+		// out after a label patch and then re-reads the new pod's IP. Patching the pod template
```
Pavani-Panakanti: can you verify if adding a label patch to pod template will update existing pod or do rollout and create new pod ? If its new pod we might not be testing the same case as issue here

Pavani-Panakanti:
> Description of changes:
The fix:

> Returns stalePodIdentifiers from deriveTargetPodsForParentCNP instead of deleting them inline.
In reconcileClusterPolicyEndpoint, removes stale identifiers from the lookup map BEFORE cleanup runs (preventing re-derivation of rules from the CPE that no longer targets them).
Explicitly clears the eBPF maps for stale identifiers by calling updateClusterPolicyBPFMaps with empty rules, which does a bulk refresh that deletes all entries and sets pod state back to DEFAULT_ALLOW.

Is this description still valid with the current changes ? 

For testing - can you also do quick test with strict mode and run all integration tests we have. Once you have these do update PR testing details. Thanks

test/integration/clusternetworkpolicy/stale_rule_cleanup_test.go
jupdec: Good catch, patching `spec.template.metadata.labels` on a Deployment does trigger a rollout, so the original test would have programmed rules on the new pod and never exercised the bug, updated test in recent commit, thank you

controllers/clusterpolicyendpoints_controller.go
jupdec: done thanks!

controllers/clusterpolicyendpoints_controller.go
jupdec: cleaned up all stale comments, thank you

test/integration/clusternetworkpolicy/stale_rule_cleanup_test.go
jupdec: makes sense, updated the test which now toggles labels on the **namespace** matching the CNPs' `SubjectNamespaces(...)` selector, not on the deployment's pod template. Deployment builder is untouched, so `Spec.Selector.MatchLabels` and `Spec.Template.ObjectMeta.Labels` never diverge and the API server accepts the patch, lmk what you think, thanks!

controllers/clusterpolicyendpoints_controller.go
jupdec: updated, thank you

controllers/clusterpolicyendpoints_controller.go
```diff
@@ -465,22 +445,62 @@ func (r *ClusterPolicyEndpointsReconciler) cleanupClusterPolicyPod(ctx context.C
...
+
+	if len(targetPodIdentifiers) == 0 {
+		r.clusterNetworkPolicyToPodIdentifierMap.Delete(parentCNP)
+		return
+	}
+	pids := make([]string, 0, len(targetPodIdentifiers))
+	for pid := range targetPodIdentifiers {
+		pids = append(pids, pid)
+	}
+	r.clusterNetworkPolicyToPodIdentifierMap.Store(parentCNP, pids)
```
Pavani-Panakanti: Can you check this seems valid to me

This stores under parentCNP, which is Spec.PolicyRef.Name (line 164). But the only reader of this map, DeriveStalePodIdentifiers, looks it up under GetParentNPNameFromPEName(cpeName) instead, which is the CPE name with everything after the last - stripped off. Two different keys for the same map.

They usually match, because the NPC creates each CPE with GenerateName: <cnpName>- and sets PolicyRef.Name: <cnpName>, so stripping the random suffix gives the CNP name back. They stop matching once the name is long enough to be truncated. The API server caps a generated name's base at 58 characters (MaxGeneratedNameLength = 63 - 5), so for any CNP name of 58 characters or more the CPE name is a truncated prefix and the stripped result is no longer the CNP name.
```
      cnp  = restrict-egress-from-payment-service-to-external-endpoints-prod   (63 chars)
      cpe  = restrict-egress-from-payment-service-to-external-endpointsABCDE

      write key (policyRef.Name)  = restrict-egress-from-payment-service-to-external-endpoints-prod.     
      read  key (stripped cpe)    = restrict-egress-from-payment-service-to-external
 ```
 When they diverge the lookup returns nothing, so stalePodIdentifiers is empty, so the scrub never runs, so cleanupClusterPolicyPod still finds the identifier's entry and takes the re-derive branch instead of the clear branch. It reprograms the Deny rules it was supposed to remove. Removing the namespace label would not restore traffic, and nothing is logged. 
A quick check: asserting that UpdateClusterPolicyEbpfMaps was called does not catch this, since both branches call it. Only the rule count tells them apart, 0 versus 1.
Second effect: the entry leaks. cleanUpClusterPolicyEndpoint computes parentCNP as GetParentNPNameFromPEName(req.Name) (line 134), so on deletion it removes a key this function never wrote, and the real one stays in the map forever.

 Make the write use the same derivation the reader already uses, so the key is byte-identical to main:
 ```
  // commitClusterPolicyEndpointState: take resourceName instead of parentCNP
  parentKey := utils.GetParentNPNameFromPEName(resourceName)
  if len(targetPodIdentifiers) == 0 {
      r.clusterNetworkPolicyToPodIdentifierMap.Delete(parentKey)
        return
    }
    r.clusterNetworkPolicyToPodIdentifierMap.Store(parentKey, pids)
 ```
 Note parentCNP is doing two unrelated jobs and they must stay separate. Line 515 compares it against each CPE's Spec.PolicyRef.Name to filter the List, so that one has to remain the real PolicyRef.Name. Only the map key should change. Renaming them (parentCNPName for the filter, parentKey for the map) would keep them from being conflated again.

File: controllers/clusterpolicyendpoints_controller.go [hunk 9]
```diff
@@ -368,69 +360,58 @@ func (r *ClusterPolicyEndpointsReconciler) deriveClusterPolicyIngressAndEgressFi
...
-		r.clusterNetworkPolicyToPodIdentifierMap.Delete(utils.GetParentNPNameFromPEName(resourceName))
-	} else {
-		r.clusterNetworkPolicyToPodIdentifierMap.Store(utils.GetParentNPNameFromPEName(resourceName), targetPodIdentifiers)
-	}
-	// Verify which pods needs to be cleaned up
 	if len(currentPods) > 0 {
 		podsToBeCleanedUp = utils.GetPodListToBeCleanedUp(currentPods, newTargetPods, podIdentifiers)
 	}
-	return newTargetPods, podIdentifiers, podsToBeCleanedUp
+	return
```
Pavani-Panakanti: empty return is a bit confusing. I understand we are using named return values but we can follow same format as rest of the repo.
Do check other places as well

Pavani-Panakanti:
Rest of the change LGTM

controllers/clusterpolicyendpoints_controller.go
jupdec: Good catch, this makes sense. Added a new test which was breaking 
```
--- FAIL: TestReconcileClusterPolicyEndpoint_StalePodIdentifiersClearedFromEbpf/truncated_CPE_name_still_triggers_stale_cleanup
    Error: Should be false
    Messages: stale identifier should be removed even when the CPE name is truncated
    Error: Should be empty, but was [{10 Deny  192.168.90.89 [] []}]
    Messages: truncated CPE name must not defeat stale cleanup &#8212; Deny rules must be cleared, not re-derived

```
which passes now with the fix [663e921](https://github.com/aws/aws-network-policy-agent/pull/605/commits/663e92156d070ad9419b9f4950c438dfb622cd1b)

```
=== RUN   TestReconcileClusterPolicyEndpoint_StalePodIdentifiersClearedFromEbpf
--- PASS: TestReconcileClusterPolicyEndpoint_StalePodIdentifiersClearedFromEbpf (0.00s)
    --- PASS: .../stale_pod_identifiers_have_eBPF_maps_cleared_on_label_removal (0.00s)
    --- PASS: .../stale_pod_with_no_eBPF_context_is_skipped,_not_errored (0.00s)
    --- PASS: .../sibling_CPE_of_the_same_CNP_does_not_keep_stale_rules_alive (0.00s)
    --- PASS: .../truncated_CPE_name_still_triggers_stale_cleanup (0.00s)
    --- PASS: .../active_pod_identifiers_keep_rules_applied_when_there_are_no_stale_identifiers (0.00s)
PASS
ok      github.com/aws/aws-network-policy-agent/controllers 0.024s

```

controllers/clusterpolicyendpoints_controller.go
jupdec: Thank you, updated in [663e921](https://github.com/aws/aws-network-policy-agent/pull/605/commits/663e92156d070ad9419b9f4950c438dfb622cd1b)

controllers/clusterpolicyendpoints_controller.go
```diff
@@ -368,69 +360,56 @@ func (r *ClusterPolicyEndpointsReconciler) deriveClusterPolicyIngressAndEgressFi
...
 	}
 
-	// Get all ClusterPolicyEndpoints for this ClusterNetworkPolicy in one call
-	parentCPEObjects := r.getClusterPolicyEndpointsOfParentCNP(ctx, parentCNP)
-	log().Infof("Parent Cluster Network Policy resource Name: %s Total Cluster Policy Endpoints for Parent CNP: Count: %d", parentCNP, len(parentCPEObjects))
+	parentCPEObjects := r.getClusterPolicyEndpointsOfParentCNP(ctx, parentCNPName)
+	log().Infof("Parent Cluster Network Policy resource Name: %s Total Cluster Policy Endpoints for Parent CNP: Count: %d", parentCNPName, len(parentCPEObjects))
 
 	if len(parentCPEObjects) == 0 {
 		podsToBeCleanedUp = append(podsToBeCleanedUp, currentPods...)
```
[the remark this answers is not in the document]
jupdec: Add a test for this case 
```
--- FAIL: TestReconcileClusterPolicyEndpoint_TransientListErrorDoesNotClearEbpf
    Error: []string{"UpdateClusterPolicyEbpfMaps", "UpdatePodStateEbpfMaps", "CreatePodStateEbpfEntryIfNotExists"} should not contain "UpdatePodStateEbpfMaps"
    Messages: pod state must not be reset on a transient List failure
    Error: Should be true
    Messages: podIdentifierToClusterPolicyEndpointMap entry must not be scrubbed on List error
    Error: Should be true
    Messages: clusterNetworkPolicyToPodIdentifierMap entry must not be deleted on List error

```
which pass now after fix [2445742](https://github.com/aws/aws-network-policy-agent/pull/605/commits/2445742735c9fe78f4a956c8acc6a3fe845390d9)

```
=== RUN   TestReconcileClusterPolicyEndpoint_StalePodIdentifiersClearedFromEbpf
    --- PASS: .../stale_pod_identifiers_have_eBPF_maps_cleared_on_label_removal (0.00s)
    --- PASS: .../stale_pod_with_no_eBPF_context_is_skipped,_not_errored (0.00s)
    --- PASS: .../sibling_CPE_of_the_same_CNP_does_not_keep_stale_rules_alive (0.00s)
    --- PASS: .../truncated_CPE_name_still_triggers_stale_cleanup (0.00s)
    --- PASS: .../active_pod_identifiers_keep_rules_applied_when_there_are_no_stale_identifiers (0.00s)
=== RUN   TestCleanUpClusterPolicyEndpoint_StalePodIdentifiersCleanedUp
    --- PASS: .../stale_identifiers_removed_from_map_during_delete_flow (0.00s)
=== RUN   TestReconcileClusterPolicyEndpoint_TransientListErrorDoesNotClearEbpf
--- PASS: TestReconcileClusterPolicyEndpoint_TransientListErrorDoesNotClearEbpf (0.00s)
PASS
ok      github.com/aws/aws-network-policy-agent/controllers 0.020s

```

controllers/clusterpolicyendpoints_controller.go
jupdec: Reverted this fix, as it's causing further issues where write key may not be unique. As this issue is an existing issue in NPA and not introduced as part of this PR, we will track it separately https://github.com/aws/aws-network-policy-agent/issues/665

jupdec:
Verified stale eBPF entries issue when namespace label is removed is fixed with this PR. Existing bug https://github.com/aws/aws-network-policy-agent/issues/665 will be tracked separately

controllers/clusterpolicyendpoints_controller.go
```diff
@@ -131,53 +131,65 @@ func (r *ClusterPolicyEndpointsReconciler) cleanUpClusterPolicyEndpoint(ctx cont
...
-	parentCNP := utils.GetParentNPNameFromPEName(req.Name)
+	parentCNPName := utils.GetParentNPNameFromPEName(req.Name)
 	resourceName := req.Name
 
-	targetPods, targetPodIdentifiers, podsToBeCleanedUp := r.deriveTargetPodsForParentCNP(ctx, parentCNP, resourceName)
+	targetPods, targetPodIdentifiers, podsToBeCleanedUp, parentCPEList, err :=
+		r.deriveTargetPodsForParentCNP(ctx, parentCNPName, resourceName)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
Pavani-Panakanti: Follow up issue to fix this as its pre existing issue https://github.com/aws/aws-network-policy-agent/issues/665

Pavani-Panakanti (approved):

PR #608, opened 2026-07-09
fix(fwruleprocessor): canonicalize CIDRs to prevent LPM key collisions

### LINKED ISSUE #609: Non-canonical CIDRs that collapse to the same LPM key are overwritten, not merged
## What happened

When two `ipBlock` egress (or ingress) rules have **different textual CIDRs that reduce to the same network after masking** — e.g. `10.161.0.0/8` and `10.0.0.0/8`, both of which are `10.0.0.0/8` once host bits are masked — the agent does not merge their L4 (port) sets. Instead the two entries collapse onto a single LPM trie key and one silently overwrites the other. Which one wins depends on Go map iteration order, so enforcement for that prefix flip-flops between reconciliations/restarts.

In a real customer incident this manifested as intermittent 5–20% TCP SYN timeouts to an STS VPC endpoint on 443, because the surviving entry sometimes carried only port 3128 (a proxy rule the operator wrote as `10.161.0.0/8` intending `10.161.0.0/16`).

## Root cause (code walkthrough)

In [`pkg/fwruleprocessor/fw_rule_processor.go`](https://github.com/aws/aws-network-policy-agent/blob/main/pkg/fwruleprocessor/fw_rule_processor.go), `ComputeMapEntriesFromEndpointRules`:

**1. The per-CIDR dedup/merge is keyed on the raw CIDR string, not the canonical (masked) network:**

```go
cidrsMap := make(map[string]EbpfFirewallRules)
...
if existingFirewallRuleInfo, ok := cidrsMap[string(firewallRule.IPCidr)]; ok {
    firewallRule.L4Info = append(firewallRule.L4Info, existingFirewallRuleInfo.L4Info...) // port union
    firewallRule.Except = append(firewallRule.Except, existingFirewallRuleInfo.Except...)
}
cidrsMap[string(firewallRule.IPCidr)] = firewallRule
```

`"10.0.0.0/8"` and `"10.161.0.0/8"` are distinct keys, so their ports are never unioned here.

**2. In the final encode loop, both strings canonicalize to the same trie key and the map write overwrites rather than merges:**

```go
for key, value := range cidrsMap {          // randomized Go map iteration order
    _, firewallMapKey, _ := net.ParseCIDR(string(key))   // 10.161.0.0/8 -> network 10.0.0.0/8
    firewallKey := utils.ComputeTrieKey(*firewallMapKey, f.enableIPv6)
    ...
    firewallMap[string(firewallKey)] = utils.ComputeTrieValue(value.L4Info, false, false)  // last write wins
}
```

`ComputeTrieKey` ([`pkg/utils/utils.go`](https://github.com/aws/aws-network-policy-agent/blob/main/pkg/utils/utils.go)) builds the key from `prefixLen` + the **masked** `n.IP`, so both inputs produce identical bytes (`prefixLen=8`, `IP=10.0.0.0`). `firewallMap[key]` is therefore written twice for the same key; the value that survives is whichever `cidrsMap` entry the range visits last.

Because Go randomizes map iteration order, the surviving port set is non-deterministic across map rebuilds (pod churn, policy edits, agent restart). A secondary contributor: `sortFirewallRulesByPrefixLength` uses `sort.Slice` (not stable), so when both rules share a prefix length the inheritance path in `checkAndDeriveL4InfoFromAnyMatchingCIDRs` is also order-dependent.

Note: the LPM lookup in the kernel is fully deterministic; the non-determinism is entirely in userspace map construction. The kernel only ever sees one value for `10.0.0.0/8`.

## What you expected to happen

Rules whose CIDRs canonicalize to the same network+prefix should have their L4 info **merged** (union of ports), deterministically. `10.161.0.0/8` + `{3128}` and `10.0.0.0/8` + `{443, ...}` should yield one `10.0.0.0/8` entry allowing `{443, ..., 3128}`.

## How to reproduce

Two egress `ipBlock`s whose CIDRs mask to the same network but carry different ports:

```yaml
egress:
- to:
  - ipBlock:
      cidr: 10.0.0.0/8        # broad allow incl. 443
  ports: [{port: 443, protocol: TCP}]
- to:
  - ipBlock:
      cidr: 10.161.0.0/8      # host bits set -> also 10.0.0.0/8; only 3128
  ports: [{port: 3128, protocol: TCP}]
```

Repeatedly rebuild the map (churn a policy-selected pod, or restart the aws-node pod) and probe an IP in `10.0.0.0/8` on 443. Connectivity intermittently succeeds/fails depending on which entry won the last write. Dumping the eBPF map shows the `10.0.0.0/8` node alternating between `{443,...}` and `{3128}`.

### Deterministic unit-test reproduction

A table test in `pkg/fwruleprocessor/fw_rule_processor_test.go` that feeds both rules and asserts the resulting `10.0.0.0/8` value contains both ports fails reliably on the current code (it hits the bad iteration within a few hundred runs) and passes with the fix.

**Before the fix:**

```
--- FAIL: TestFWRuleProcessor_NonCanonicalCIDRCollision
    iteration 3: 10.0.0.0/8 entry lost a port; got ports=[443] (want both 443 and 3128).
    Non-canonical CIDR 10.161.0.0/8 collided with 10.0.0.0/8 and overwrote its L4 info.
FAIL
```

**After the fix:**

```
--- PASS: TestFWRuleProcessor_NonCanonicalCIDRCollision (0.12s)
ok  github.com/aws/aws-network-policy-agent/pkg/fwruleprocessor
```

The full `pkg/fwruleprocessor` suite continues to pass and `go vet` is clean.

## Suggested fix

Canonicalize the CIDR before it is used as the dedup/merge key, so masking happens once, up front:

```go
if _, ipNet, err := net.ParseCIDR(string(firewallRule.IPCidr)); err == nil {
    firewallRule.IPCidr = v1alpha1.NetworkAddress(ipNet.String())
}
```

This makes the existing port-union path fire for host-bits-set duplicates and removes the colliding double-write in the encode loop. Additionally, switch `sortFirewallRulesByPrefixLength` to `sort.SliceStable` to remove the secondary ordering dependency.

## Impact

Silent, intermittent allow/deny for any prefix a user expresses with host bits set (a common typo, e.g. `/8` intended as `/16`). Hard to diagnose because it is non-deterministic, passes basic connectivity checks intermittently, and produces no error in the agent logs.

## Related

- PR #519 — rule-combination tests for same-string CIDRs (does not cover canonicalization collisions).
- Distinct from the VPC CNI v1.21.1 "map pointer overwrite" fix (that was the whole map handle; this is a per-key value overwrite).

git4example:
Fixes #609

## What does this PR do?

Canonicalizes each `ipBlock` CIDR (masks off host bits) before it is used as
the `cidrsMap` key in `FirewallRuleProcessor.ComputeMapEntriesFromEndpointRules`,
so rules that reduce to the same network are merged deterministically instead of
colliding and overwriting one another.

## Why?

Two rules whose CIDRs canonicalize to the same network — e.g. `10.0.0.0/8` and
`10.161.0.0/8` (a `/8` mask makes the `.161` octet meaningless, so both are
`10.0.0.0/8`) — were keyed separately by their raw string. Their L4 port sets
were never unioned, and because both encode to the identical LPM trie key, the
final `range cidrsMap` write silently overwrote one with the other in Go's
randomized map-iteration order.

The result was non-deterministic port enforcement across reconciliations. In a
real incident this surfaced as intermittent (~5–20%) TCP SYN timeouts to an STS
VPC endpoint on port 443, whenever the surviving `10.0.0.0/8` entry happened to
carry only the proxy port 3128.

The kernel LPM lookup is deterministic; the bug is entirely in userspace map
construction.

## Changes

- `pkg/fwruleprocessor/fw_rule_processor.go`
  - Canonicalize the CIDR (`net.ParseCIDR(...).String()`) before it is used as
    the `cidrsMap` key, so the existing port-union path handles host-bits-set
    duplicates.
  - `sort.Slice` → `sort.SliceStable` in `sortFirewallRulesByPrefixLength` to
    remove the secondary ordering dependency in the L4-info inheritance path.
- `pkg/fwruleprocessor/fw_rule_processor_test.go`
  - Add `TestFWRuleProcessor_NonCanonicalCIDRCollision`, which asserts the merged
    `10.0.0.0/8` entry always retains both ports across many rebuilds.

## Testing

`TestFWRuleProcessor_NonCanonicalCIDRCollision` fails on the previous code and
passes with this change:

```
# before
--- FAIL: TestFWRuleProcessor_NonCanonicalCIDRCollision
    iteration 3: 10.0.0.0/8 entry lost a port; got ports=[443] (want both 443 and 3128).

# after
--- PASS: TestFWRuleProcessor_NonCanonicalCIDRCollision (0.12s)
ok  github.com/aws/aws-network-policy-agent/pkg/fwruleprocessor
```

Full `pkg/fwruleprocessor` suite passes and `go vet ./pkg/fwruleprocessor/` is
clean.

## Related

- PR #519 (rule-combination tests) covers same-string CIDRs but not
  canonicalization collisions.
- Distinct from the VPC CNI v1.21.1 "map pointer overwrite" fix (whole map
  handle vs. this per-key value overwrite).

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.


emilyhuaa:
Thanks @git4example for the fix. Confirmed the repro locally, `TestFWRuleProcessor_NonCanonicalCIDRCollision` fails on main and passes with the change. Approach looks good.

Couple of things to consider:

We have to make the same change in `ComputeClusterPolicyMapEntriesFromEndpointRules` too. It uses `f.normalizeCIDR` which only appends the host mask and doesn't mask host bits, so the same collision applies to the CNP path. I ran a quick test on your branch and it reproduces there.

The except CIDR path also skips canonicalization. If a rule allows `10.0.0.0/8:443` with `except: [10.161.0.0/8]`, the except gets stored under its raw string, misses the guard on line 144, and ends up adding a DENY at the same LPM key as the allow. Saw the allow get overwritten on ~440/500 iterations in a local test.

Minor: `sortByCIDRLength` in the CNP path could also use `sort.SliceStable` for consistency.

git4example:
Thanks for the thorough review @emilyhuaa! All three points addressed in dc11277:

1. **CNP path** — `normalizeCIDR` now masks host bits via `net.ParseCIDR`, same approach as the pod-scoped path.
2. **Except CIDRs** — canonicalized before the `cidrsMap` lookup so a non-canonical except can't bypass the guard and overwrite the allow entry.
3. **`sortByCIDRLength`** — switched to `sort.SliceStable`.

Added two new tests (`TestClusterPolicy_NonCanonicalCIDRCollision` and `TestFWRuleProcessor_ExceptCIDRCanonicalization`) — both run 500 iterations and pass consistently.

emilyhuaa:
Confirmed copilot comment locally. This isn't introduced by the PR. Pre-PR, an except CIDR with host bits (e.g. rule `10.0.0.0/8:443` + `except: [10.161.0.0/8]`) would silently DENY-overwrite the allow at the shared LPM key (repro'd 440/500 iters on main). This PR improves that to "silently ignored except" via the canonicalized lookup, which is the state Copilot flagged.

Not blocking. Couple things for a follow-up:
1. Warn log when the canonicalized except collides with an existing allow, so this is diagnosable from logs.
2. Per K8s NetworkPolicy semantics, `except` should still subtract from the enclosing allow. Covers-whole-prefix case currently leaves the allow intact.

Otherwise lgtm, tests reproduce the underlying bugs cleanly on main.

emilyhuaa (approved):
lgtm

PR #610, opened 2026-07-09
Conntrack port reuse GC race fix

Pavani-Panakanti:
## Problem

Pods behind a NetworkPolicy periodically lose connectivity: a short-lived flow's return traffic is `Verdict DENY`'d even though policy allows it, then recovers on its own. 

The cause is a race in the userspace conntrack cleanup

1. Cleanup snapshots the kernel conntrack table.
2. It then iterates the eBPF `aws_conntrack_map` and deletes every entry not in
   that snapshot.

Between the two steps a short-lived connection can close and its ephemeral source port be reused by a new connection. The new flow re-inserts a live map entry that was not in the snapshot, so cleanup deletes a **live** entry. The next return packet finds no conntrack entry and is denied; the map self-heals on a later packet, so the freeze is transient.

## Fix

Let cleanup distinguish "stale" from "just reused" without a second kernel snapshot: stamp every datapath hit with a monotonic timestamp, and re-check it before deleting.

- **Datapath (eBPF C, v4/v6 ingress+egress + events):** add `last_seen` to `conntrack_value`, set via `bpf_ktime_get_ns()` on every flow hit (new flow, forward-flow match, reverse-flow match). Each site writes through a map value pointer the code already holds — no extra map lookup or update.
- **`conntrack_value` grows 1B → 16B:** `__u64 val` + `__u64 last_seen`. `val` was widened from `__u8` to `__u64` rather than adding `__u8 _pad[7]`: an 8-byte field must sit at an 8-byte-aligned offset, so either form is 16B, and the wider field avoids explicit padding. `ConntrackVal` in `pkg/utils/utils.go` mirrors the layout.
- **Userspace GC:** for each delete candidate (absent from the kernel snapshot), re-read the live map value; if `last_seen >= gcStart` the flow was touched during this cycle → **skip deletion** (the reused-port case). The re-read **fails safe** — on error the entry is kept. The decision is isolated in the pure predicate `entryActiveFromRead(val, readErr, gcStart)`.
- **CLI:** `aws-eks-na-cli ebpf dump-maps` prints `Last Seen (ns)`.

`gcStart` is read once per cycle via `utils.KtimeGetNs()` (`clock_gettime(CLOCK_MONOTONIC)`), the userspace counterpart to the kernel's `bpf_ktime_get_ns()`. It is shared by the GC and the CLI so the two cannot drift onto different clocks.

## Design notes

- **Clock domains match.** The predicate compares a kernel-written `last_seen` against a userspace-read `gcStart`. Both read the same kernel CLOCK_MONOTONIC (`bpf_ktime_get_ns` ↔ `clock_gettime(CLOCK_MONOTONIC)`), so the comparison is meaningful.
- **Residual TOCTOU is bounded, not eliminated.** A flow can still be reused in the gap between the re-read and the delete, and an entry reused inside that gap is still deleted. The window shrinks from the full snapshot-to-iterate span to a single syscall, so the exposure is orders of magnitude smaller, but it is not zero. 
- **Map layout is forward/backward compatible.** On any version switch the events `.o` differs, and a change to that binary sets the events/ingress/egress update flags together, so both recovery paths are skipped and the map is recreated at the new `value_size` — old and new layouts are never mixed. Known gap: if the agent is killed after the new binaries are copied but before the map is recreated, the next start compares packaged-vs-host binaries, finds them equal, and recovers the old map against new expectations. Hardening the startup check to compare live map `value_size` (not just binary bytes) is tracked as a follow-up.

## Performance

**Per-packet datapath cost.** The stamp is one `bpf_ktime_get_ns()` call plus an
8-byte store; there is no map lookup and no syscall (TC BPF runs in-kernel).
Measured with `BPF_PROG_TEST_RUN` (1M repeats):

| | ns/packet |
|---|---|
| Added cost per matched flow | **~34 ns** |

Cost is O(1) per packet 

**GC cleanup cost vs batch size.** The GC adds exactly one `GetMapEntry` (`BPF_MAP_LOOKUP_ELEM`) syscall per deleted entry. 

The re-read roughly doubles GC wall time and CPU (one added syscall per delete), scaling linearly with entries deleted. GC is a tight syscall loop, so CPU tracks wall time. This runs in a background goroutine off the datapath.

**Memory.** The value widening 1B→16B raises the map's kernel memlock by a fixed **~4 MB** (49 MB → 53 MB, preallocated for `max_entries=524288`),

## Testing

**Unit** (`pkg/ebpf/conntrack/conntrack_client_test.go`):
- `TestConntrackValLayout` — struct is 16B, `LastSeen` at offset 8.
- `TestKtimeGetNs` (`pkg/utils/utils_test.go`) — returns nonzero, monotonic non-decreasing.
- `TestEntryActiveFromRead` — refreshed→keep, stale→delete, `==gcStart`→keep,
  re-read-fail→keep (fail-safe), zero-last_seen→delete.
- All `pkg/...` unit tests pass; `go build ./...`, `go vet`, `gofmt` clean.

**End-to-end race repro (new soak test)
- Stock NPA → **FAILS**: return-flow `Verdict DENY` observed.
- Patched NPA → **PASSES**: **0 DENY** over the run. The per-entry
  `Conntrack cleanup Skip (in use)` line that shows the skip engaging is logged at
  debug level, so it is visible only with `--log-level=debug`.
- Run on both IPv4 and IPv6 clusters: 30 min, 0 DENY, 100% poll coverage on each.

**Upgrade/downgrade (map recreation on version switch):**
- Upgrade to patched: conntrack map recreated at `value 16B` (`bpftool`), CLI
  dump shows `Last Seen (ns)` populated.
- Downgrade to a lower version: map recreated at `value 1B`; agent healthy.
- Both directions were exercised on matched CNI/NPA image pairs.

cli output:

```
Conntrack Key : Source IP - 192.168.71.164 Source port - 41912 Dest IP - 10.100.4.247 Dest port - 80 Protocol - 6 Owner IP - 192.168.71.164 Ifindex - 6
  Value :
  Conntrack Val -  2
  Last Seen (ns) -  21960846150943  (1.153s ago)
  *******************************
Conntrack Key : Source IP - 192.168.71.164 Source port - 46227 Dest IP - 10.100.0.10 Dest port - 53 Protocol - 17 Owner IP - 192.168.71.164 Ifindex - 6
  Value :
  Conntrack Val -  2
  Last Seen (ns) -  21920725006909  (41.274s ago)
  *******************************
Conntrack Key : Source IP - 192.168.71.164 Source port - 52914 Dest IP - 10.100.0.10 Dest port - 53 Protocol - 17 Owner IP - 192.168.71.164 Ifindex - 6
  Value :
  Conntrack Val -  2
  Last Seen (ns) -  21862542824821  (1m9.697s ago)
  *******************************
  ```

### CODE REVIEW

File: pkg/clihelper/show.go [hunk 1]
```diff
@@ -226,6 +226,7 @@ func MapWalk(mapID int, mapNamePrefix string) error {
 					fmt.Println(retrievedKey)
 					fmt.Println("Value : ")
 					fmt.Println("Conntrack Val - ", iterValue.Value)
+					fmt.Println("Last Seen (ns) - ", iterValue.LastSeen)
```
viveksb007: can you add the output to the PR details?

pkg/ebpf/c/tc.v4egress.bpf.c
```diff
@@ -446,6 +449,7 @@ int handle_egress(struct __sk_buff *skb)
 		if (flow_val != NULL) {
 			// If the pod state matches, allow the packet
 			if (flow_val->val == ct_pod_state_val) {
+				flow_val->last_seen = bpf_ktime_get_ns();
```
viveksb007, line 455: these values are not being written to bpf map right?

we are changing the value on stack and then returning, so for fast path of conntrack bypass, don't think we are updating the last_seen timestamp

pkg/ebpf/c/tc.v4egress.bpf.c
```diff
@@ -473,6 +477,7 @@ int handle_egress(struct __sk_buff *skb)
 		reverse_flow_val = bpf_map_lookup_elem(&aws_conntrack_map, &reverse_flow_key);
 
 		if (reverse_flow_val != NULL) {
+			reverse_flow_val->last_seen = bpf_ktime_get_ns();
```
viveksb007, line 484: same as above this value won't be written into ebpf map

pkg/ebpf/c/tc.v4egress.bpf.c
```diff
@@ -90,6 +90,8 @@ struct conntrack_key {
 
 struct conntrack_value {
    __u8 val; // 0 => default-allow, 1 => policies-applied
+   __u8 _pad[7];
```
viveksb007: can we avoid this padding?

i remember there is some alignment issue. if we have to increase the storage value, then maybe convert `u8` to `u64` val for better readability

pkg/ebpf/conntrack/conntrack_client.go
```diff
@@ -17,6 +17,41 @@ var (
...
+	}
+	return cur.LastSeen >= gcStart
+}
+
+// isEntryActive re-reads the live BPF map entry and returns true if it has
+// been touched by the datapath since gcStart (i.e., last_seen >= gcStart).
+// On read failure the entry is assumed active (fail-safe: don't delete).
+func (c *conntrackClient) isEntryActive(key unsafe.Pointer, gcStart uint64) bool {
+	var cur utils.ConntrackVal
+	err := c.conntrackMap.GetMapEntry(uintptr(key), uintptr(unsafe.Pointer(&cur)))
```
viveksb007, line 38: this is probably same class of issue from ebpf-sdk related to uintptr and syscall to be in same expression

pkg/ebpf/conntrack/conntrack_client.go
```diff
@@ -17,6 +17,41 @@ var (
...
+// unit testing without a live BPF map. An entry is "active" (must NOT be
+// deleted) if the re-read failed (fail-safe) or the datapath refreshed its
+// last_seen at or after gcStart (i.e., the flow was touched during this GC
+// cycle — the port-reuse race case).
+func entryActiveFromRead(cur utils.ConntrackVal, readErr error, gcStart uint64) bool {
+	if readErr != nil {
+		return true
+	}
+	return cur.LastSeen >= gcStart
+}
```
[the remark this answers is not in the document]
Pavani-Panakanti: If read fails to check if entry is up for deletion - fail safe and keep the entry

pkg/ebpf/c/tc.v4egress.bpf.c
Pavani-Panakanti: The stamp does persist to the map. This is the in-place update path for a hash map value from a BPF program, not writing to local copy.
In BPF program context, bpf_map_lookup_elem returns a pointer directly into the map's value memory, not a copy. From the kernel BPF hash-map documentation (https://docs.kernel.org/bpf/map_hash.html):
> "Hash entries can be retrieved using the bpf_map_lookup_elem() helper. This helper returns a pointer to the value associated with key, or NULL if no entry was found."
The same doc's usage example writes through that pointer to mutate the live entry in place, which is exactly pattern used:
 ```
  struct value *value = bpf_map_lookup_elem(&packet_stats, &key);
  if (value) {
        __sync_fetch_and_add(&value->packets, 1);   // mutates the map entry in place
  }
  ```
So flow_val = bpf_map_lookup_elem(...); flow_val->last_seen = bpf_ktime_get_ns() updates the map entry itself. No bpf_map_update_elem is needed, and adding one would be more expensive. 
Adding this detail - The userspace syscall form (bpf_map_lookup_elem(int fd, const void *key, void *value)) which we use in GC cleanup copies into buffer. 

I have done some extensive tests to validate this behavior. Added a comment as well so its clear for readers

pkg/ebpf/c/tc.v4egress.bpf.c
Pavani-Panakanti: makes sense. Updated

pkg/ebpf/conntrack/conntrack_client.go
Pavani-Panakanti: Yeah we need to update all ebpf sdk api calls for this. I will handle all these in separate PR's. First make changes in sdk and then update all calls here

test/integration/soak/port-reuse/port_reuse_suite_test.go
```diff
@@ -0,0 +1,45 @@
+//go:build soak
```
yash97, line 1: F1 (should fix) — soak test counts are absolute, not deltas. logCount greps the entire on-node NPA log since agent start. Three consequences:

  - Any pre-existing Verdict DENY line on that node — from a prior test, another namespace's workload, anything — trips Expect(deny).To(Equal(0)) instantly: false failure on any non-pristine node.
  - Pre-existing Conntrack cleanup Delete lines satisfy the liveness gate without cleanup actually running now.
  - Log rotation mid-soak makes counts go backwards.

  The validation runs passed because the cluster was fresh. Fix: snapshot all three counts after warmup and assert on deltas; ideally also scope the DENY grep to the test pods' IPs (the line format includes source/dest IPs) so a same-node
  bystander can't fail the test. This matters because a soak test that false-fails on shared clusters stops getting run.

pkg/clihelper/show.go
Pavani-Panakanti: added

yash97:
The upgrade/downgrade validation is solid. Can we make it permanent? A small generic helper that checks each global map's live spec (value_size, key_size, max_entries) against what we expect, ideally with value_size == unsafe.Sizeof(ConntrackVal{}) so C, kernel, and Go can't drift. Called once from the existing upgrade suite.

yash97:
Also can we run  the integration test and soak test multiple times ideally 5 or 10 times to make sure there is no flakiness.

yash97:
One more check did we run soak test integration test in ipv4 and ipv6 cluster multiple times? I see we made changes in test but lets please ensure ipv6 vertical is covered too.

Pavani-Panakanti:
> The upgrade/downgrade validation is solid. Can we make it permanent? A small generic helper that checks each global map's live spec (value_size, key_size, max_entries) against what we expect, ideally with value_size == unsafe.Sizeof(ConntrackVal{}) so C, kernel, and Go can't drift. Called once from the existing upgrade suite.

This will be a generic check we can add. Its not related to upgrades. Its a good check to have. I will add a follow up task for this one. Its not related to changes in this PR its general check we can add for all maps we have 

> Also can we run the integration test and soak test multiple times ideally 5 or 10 times to make sure there is no flakiness.

Yes ran tests multiple times and they are reliable - no flakiness. 

> One more check did we run soak test integration test in ipv4 and ipv6 cluster multiple times? I see we made changes in test but lets please ensure ipv6 vertical is covered too.

I made some changes to keep the test generic. Earlier it was testing just this race but want to keep it generic. Run port reuse workload for 30 min and check for any DENY's so that we catch any issues with this kind of workload in future and not just this specific race condition. No other issues 
I have skipped ipv6 intentionally. Its a generic workload specific soak test for catching any behavior regressions that runs longer for 30 min. Adding both ipv4 and v6 will take longer to add suite in automations. We can test individual diff between ipv4 and ipv6 as integration tests with shorter times.

pkg/clihelper/show.go
```diff
@@ -20,6 +21,32 @@ type PodState struct {
 	State uint8
 }
 
+// ktimeGetNs returns the current CLOCK_MONOTONIC time in nanoseconds — the same
+// clock the datapath stamps last_seen with via bpf_ktime_get_ns(), so the two
+// can be subtracted directly.
```
yash97: Non blocking comment:- But seems like we do not need comments on every function. We can keep them concise and only keep them where there are needed to provide some contract or not easily understandable decision.

pkg/clihelper/show.go
```diff
@@ -4,6 +4,7 @@ import (
 	"errors"
```
yash97, line 4: fyi with scema change this is the risk still open 

on pkg/ebpf/bpf_client.go:203, at:
```
   if eventsUpdateRequired || (!isConntrackMapPresent || !isPolicyEventsMapPresent) {
```

Problem: This condition can miss the map migration after an interrupted upgrade. InstallBPFBinaries copies the new binaries before LoadBpfFileWithCustomData recreates the conntrack map. If the agent exits in between, the next startup sees that the packaged and host binaries already match, so eventsUpdateRequired is false. Because the old map is still pinned, isConntrackMapPresent is also true, and this branch is skipped. The new agent can then recover the old one-byte map even though ConntrackVal and the new TC programs expect a 16-byte value. Userspace reads will see LastSeen as zero, and loading a new TC program may fail verifier checks when it accesses the field at offset 8. An interrupted downgrade has the inverse risk: old userspace may use a one-byte destination while reading from the 16-byte map.
   Suggested approach: Before reusing the pinned map, read its live map information and validate its type, key size, and value size. Treat any schema mismatch—especially:
 
```
   > mapInfo.ValueSize != uint32(unsafe.Sizeof(utils.ConntrackVal{}))
  
```
 
as a reason to recreate the map, regardless of whether the binaries currently match. Please also add a restart test for the state where the new binary has already been copied but the old map is still pinned. The current clean upgrade/downgrade test does not exercise this interruption point.

pkg/ebpf/conntrack/conntrack_client.go
```diff
@@ -188,12 +229,20 @@ func (c *conntrackClient) CleanupConntrackMap() {
 			newKey.Ifindex = 0 // strip for 5-tuple comparison
 			_, ok := kernelConntrackV4Cache[newKey]
 			if !ok {
-				// Delete the entry in local cache since kernel entry is still missing so expired case
+				// Candidate for deletion — absent from kernel snapshot.
+				// Re-read the live BPF entry to check if it was touched since GC started
+				// (handles port-reuse race: a new connection refreshes last_seen).
 				expiredFlow := localConntrackEntry
+				if c.isEntryActive(unsafe.Pointer(&expiredFlow), gcStart) {
+					log().Infof("Conntrack cleanup Skip (in use) - Source IP - %s Source port - %d Dest IP - %s Dest port - %d Protocol - %d Owner IP - %s Ifindex - %d",
```
viveksb007: This logs one Info line per skipped entry per GC cycle (same in the v6 path below). Under the exact churn workload this fix targets — sustained port reuse — a single cycle could skip many entries. It's consistent with the existing per-delete Info logging, so not blocking, but consider Debug for the per-entry line, or a single per-cycle skip count at Info, to keep the NPA log usable during the scenario this is designed for.

pkg/ebpf/conntrack/conntrack_client.go
```diff
@@ -17,6 +17,41 @@ var (
 	CONNTRACK_MAP_PIN_PATH = "/sys/fs/bpf/globals/aws/maps/global_aws_conntrack_map"
 )
 
+// ktimeGetNs returns the current CLOCK_MONOTONIC time in nanoseconds,
+// matching bpf_ktime_get_ns() used by the datapath to stamp last_seen.
+func ktimeGetNs() uint64 {
```
viveksb007: `ktimeGetNs` is now defined in both `pkg/clihelper` and `pkg/ebpf/conntrack`, with divergent error handling (silent return 0 vs `Errorf`). Consider a shared helper (e.g. in `pkg/utils`) so the two copies can't drift apart. Non-blocking.

pkg/ebpf/conntrack/conntrack_client.go
```diff
@@ -17,6 +17,41 @@ var (
...
+	if readErr != nil {
+		return true
+	}
+	return cur.LastSeen >= gcStart
+}
+
+// isEntryActive re-reads the live BPF map entry and returns true if it has
+// been touched by the datapath since gcStart (i.e., last_seen >= gcStart).
+// On read failure the entry is assumed active (fail-safe: don't delete).
+func (c *conntrackClient) isEntryActive(key unsafe.Pointer, gcStart uint64) bool {
```
viveksb007, line 33: Worth a one-line comment acknowledging the concurrent-write race here: the re-read copies the 16-byte value while the datapath may be storing `last_seen` into it. Aligned u64 stores are effectively atomic on x86-64/arm64 and a torn read worst-case produces one wrong GC decision that self-heals next cycle, so it's fine — but future readers will wonder whether it was considered.

viveksb007 (commented):
Reviewed the datapath stamping, GC re-check logic, and upgrade path. The core fix looks correct:

- `gcStart` is read before the kernel conntrack snapshot in both v4/v6 paths, so any flow created after the snapshot necessarily has `last_seen > gcStart` and is kept. The residual TOCTOU between re-read and delete fails toward keeping entries — the safe direction.
- Clock domains match: `bpf_ktime_get_ns()` and `clock_gettime(CLOCK_MONOTONIC)` read the same clock.
- Verified in aws-ebpf-sdk-go v1.0.15 (`pkg/maps/loader.go`) that map creation unpins and replaces an existing pin, and `eventsUpdateRequired` forces that path on upgrade — so the 1B→16B pinned map is recreated, never reused with a mismatched layout. Old attached programs hold their own reference to the old map, so there's no mixed-layout access during the transition.
- Every error path (clock read failure, map re-read failure) fails toward keeping entries, and kept-but-dead entries are collected on the next cycle since the cache is re-hydrated.
- The layout-drift sentinel tests and the soak test's liveness/coverage gates (a run that was mostly blind fails instead of passing) are exactly the right kind of tests for a race fix.

One nit not tied to a line: the PR description still describes the layout as `__u8 val` + `__u8 _pad[7]` + `__u64 last_seen`, but f120e45 changed it to `__u64 val` + `__u64 last_seen`. Worth updating the description so the merged record matches the code.

A few minor comments inline.

pkg/ebpf/c/tc.v4egress.bpf.c
viveksb007: this was interesting to learn. thanks

pkg/clihelper/show.go
Pavani-Panakanti: Reviewed all newly added comments. All provide good info and context for future maintenance. I will keep them all but made them concise wherever possible

pkg/ebpf/conntrack/conntrack_client.go
Pavani-Panakanti: updated to use a util func

pkg/clihelper/show.go
Pavani-Panakanti: The window is very narrow where np agent has to crash for this issue to trigger but very valid issue and we should address it. This should be addressed for all our bpf maps (ingress/egress/events/conntrack). Considering its pre existing and requires very narrow crash window we can handle this in a separate PR covering all maps
```
What change would be:
current: updateRequired = binaryChanged
new:     updateRequired = binaryChanged || schemaMismatch
```

pkg/ebpf/c/tc.v4egress.bpf.c
```diff
@@ -89,7 +89,8 @@ struct conntrack_key {
 };
 
 struct conntrack_value {
-   __u8 val; // 0 => default-allow, 1 => policies-applied
+   __u64 val; // pod state, set via GET_CT_VAL
```
viveksb007: not fully clear with the comment `set via GET_CT_VAL`, what does it mean?

pkg/ebpf/c/tc.v4egress.bpf.c
```diff
@@ -89,7 +89,8 @@ struct conntrack_key {
...
+   __u64 last_seen; // bpf_ktime_get_ns() of the last datapath hit; used by userspace GC to skip reused flows
```
viveksb007: nit - can simply this comment to tell what this value signifies, we don't need to mention kernel fn which populates it in the comment itself

viveksb007 (commented):
logic lgtm. if we can make the LLM comments brief to actual code understanding and remove the session specific thing that we are trying to solve port reuse GC race, that will be great.

pkg/ebpf/conntrack/conntrack_client.go
```diff
@@ -111,6 +136,14 @@ func (c *conntrackClient) CleanupConntrackMap() {
...
 		// Conntrack table is already hydrated from previous run
 		// So read from kernel conntrack table
+		gcStart, err := utils.KtimeGetNs()
+		if err != nil {
+			// Without a reference time every entry would look stale, so skip the
+			// cycle rather than risk deleting active entries.
+			log().Errorf("clock_gettime(CLOCK_MONOTONIC) failed, GC skipping eviction this cycle: %v", err)
+			c.hydratelocalConntrack = true
+			return
+		}
```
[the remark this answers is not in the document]
Pavani-Panakanti: Discussed here https://github.com/aws/aws-network-policy-agent/pull/610/changes#r3714122055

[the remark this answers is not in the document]
viveksb007: can you open a GH issue and mention these details or point to this discussion? we can pick it separately

viveksb007 (approved):

pkg/ebpf/conntrack/conntrack_client.go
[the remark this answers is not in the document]
Pavani-Panakanti: Open github issue to track this https://github.com/aws/aws-network-policy-agent/issues/651

PR #614, opened 2026-07-12
test: fix flaky IPBlock-Except spec (premature pod-log read)

yash97:
## What

Fix the intermittently-failing `IPBlock Except — CIDR and Except overlap`
integration spec (`test/integration/policy/except_block_test.go`), which fails
~1-in-6-to-12 runs with:

```
Expected deny to server on port 3307 but got allow
```

## Root cause — a timing race in the test, NOT in enforcement

The client pod sleeps, then probes the server on **3306 (allowed)** and **3307
(denied)** with `nc -z -w1`; the `It` block sleeps a **fixed 30s** and reads the
pod log **once**.

A denied port is **silently dropped** by the policy datapath (no RST), so
`nc -z -w1` on 3307 must block its **full 1-second timeout** before printing
`CLOSE-3307`, whereas the allowed 3306 probe prints `OPEN-3306` in ~0ms. The two
sleeps also start on offset clocks (the reader's `It` only begins after the pod
is detected `Running`). When the reader's fixed sleep expired inside that ~1s
window, it saw `OPEN-3306` but **not yet** `CLOSE-3307`, and `processIPBlockLogs`
maps a missing `CLOSE-3307` to the misleading `"got allow"`.

### Enforcement was correct — verified at failure time

Reproduced on an IPv6 EKS cluster with failure-time datapath capture:

- **NPA per-packet event log:** Dest Port 3307 = **37 × `Verdict: DENY`, 0 ACCEPT**
  (Tier `NETWORK_POLICY`); the original failing packet is logged `DENY`.
- **eBPF `global_aws_conntrack_map`:** no 3307 entry → no conntrack fast-path
  allow.
- **LPM egress map:** server `/64` → TCP 3306 only; pod_state = enforce.

The 3307 packet was dropped; the test just read the log ~0.36s too early.

## Fix

Eliminate the timing race entirely: drop the fixed sleeps and one-shot log read,
and probe connectivity directly with `Eventually`/`Consistently`.

- `tcpProbe` execs `nc -z` from the client pod on demand and returns the `OPEN`/
  `CLOSE` verdict, so the test observes live connectivity instead of parsing a log
  the pod wrote once.
- Enforcement (deny on the excepted port) is asserted with `Eventually(...).Should
  (Equal("CLOSE"))`, and it is checked first so a converged deny proves enforcement
  is active before the allow probes run.
- The allowed port and the endpoint outside the excepted prefix use `Eventually
  (...).Should(Equal("OPEN"))`.
- A final `Consistently(...).Should(Equal("CLOSE"))` confirms the deny did not flap
  while the allow probes ran.

No fixed sleeps and no log parsing remain, so the read can no longer land inside
the probe's completion window.

## Testing

Ran the previously-flaky spec **10× on an IPv6 EKS cluster (standard mode) → 10/10
PASS** (`--focus "CIDR and Except overlap" --ip-family=IPv6`). Pre-fix the same
spec failed within a handful of runs.

The IPv4 path is behavior-equivalent (same sleeps; the race is IP-family-agnostic).

### DIFF
--- a/test/integration/policy/except_block_test.go
+++ b/test/integration/policy/except_block_test.go
@@ -16,6 +16,12 @@ import (
 	network "k8s.io/api/networking/v1"
 )
 
+const (
+	enforcementTimeout = 90 * time.Second // wait for policy to be programmed into the datapath
+	probeTimeout       = 30 * time.Second // allow probes once enforcement is active
+	probeInterval      = 3 * time.Second
+)
+
 func printNetworkPolicyYAML(np *network.NetworkPolicy) {
 	bytes, err := yaml.Marshal(np)
 	if err != nil {
@@ -83,9 +89,9 @@ var _ = Describe("IPBlock Except Test Cases", func() {
 		By("Deploying a sample TCP server on ports 3306 & 3307", func() {
 			err := fw.NamespaceManager.CreateNamespace(ctx, serverNamespace)
 			Expect(err).ToNot(HaveOccurred())
-			// listening on two ports
+			// Per-port re-listen loop: busybox nc -l exits after one connection.
 			cmd := fmt.Sprintf(
-				"while true; do nc -l -p %d & nc -l -p %d & wait; done",
+				"while true; do nc -l -p %d; done & while true; do nc -l -p %d; done & wait",
 				allowPort, blockPort,
 			)
 			srv := manifest.NewBusyBoxContainerBuilder().
@@ -107,26 +113,12 @@ var _ = Describe("IPBlock Except Test Cases", func() {
 		})
 	})
 
+	// deployClient creates an idle pod we exec probes into via Eventually.
 	deployClient := func(clientName string) *v1.Pod {
-		cfg := ipFamilyConfigForIP(serverIP)
-		// A denied port is silently dropped, so its `nc -w1` blocks 1s before
-		// printing CLOSE- while the allowed port prints instantly. The It block
-		// must read only after this whole probe phase completes.
-		script := fmt.Sprintf(
-			`sleep 10;
-	nc -z -w1 %s %d && echo "OPEN-%d" || echo "CLOSE-%d";
-	nc -z -w1 %s %d && echo "OPEN-%d" || echo "CLOSE-%d";
-	nc -z -w2 %s %d && echo "OPEN-EXT" || echo "CLOSE-EXT";
-	sleep 1000 `,
-			serverIP, allowPort, allowPort, allowPort,
-			serverIP, blockPort, blockPort, blockPort,
-			cfg.extProbeIP, 53,
-		)
-
 		ctnr := manifest.NewBusyBoxContainerBuilder().
 			ImageRepository(fw.Options.TestImageRegistry).
 			Command([]string{"/bin/sh", "-c"}).
-			Args([]string{script}).
+			Args([]string{"sleep 1000000"}).
 			Build()
 
 		clientPod = manifest.NewDefaultPodBuilder().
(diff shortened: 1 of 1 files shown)


test/integration/policy/except_block_test.go
```diff
@@ -172,7 +175,8 @@ var _ = Describe("IPBlock Except Test Cases", func() {
 		})
 
 		It("should allow on server prefix and 3306 port, deny on rest server-prefix ports, allow all on rest of endpoints", func() {
-			time.Sleep(30 * time.Second)
+			// > client sleep(10) + probe phase, so all probe results are logged first.
+			time.Sleep(15 * time.Second)
```
viveksb007: can you provide some clarity on the changes?

flow of test
- deploy serverPod in beforeEach section
- then as part of this test, (its beforeEach deploys a NP and creates a client)
- in this test, we have moved sleep from 30 to 15s

I understand, it should speed up the test but not getting where is flakiness. can you explain the flake part

yash97: you see, in before each we deploy pod and sleep for 30 seconds and then immediately deploy pod log and sleep for 30 seconds. Now test pods need some 1 or 2 second buffer, before it completes execution of probes. but right now difference between checker and execution is very less. This test keeps floor of enforcement at 10 seconds and buffer of 5 seconds to execute checks.

viveksb007: discussed offline, this PR is increasing the delta b/w client pods getting deployed, sleeping for 10s, then making a call and logging.

and this test is waiting for 15s and then checking the logs. we should move away from this timing based test and use exec in client pod to make the call and do validation.

something like, deploy client pod, wait for it to come up and do `curl/nc` via exec to make assertions. currently test is flaky because of time and assumptions that other parts of env like k8s control plane, etc are fine.

we have some tests that use ValidateConnection construct, example https://github.com/aws/aws-network-policy-agent/blob/ef3b0748bb0812ef4aee804967cf025cec245fdb/test/integration/clusternetworkpolicy/domain_policy_test.go#L155
we should be able to migrate these tests to that pattern

test/integration/policy/except_block_test.go
```diff
@@ -172,49 +167,65 @@ var _ = Describe("IPBlock Except Test Cases", func() {
...
+		fmt.Sprintf(`nc -z -w2 %s %d 2>/dev/null && echo "OPEN" || echo "CLOSE"`, host, port),
 	}
-	if !gotAllowExt {
-		return fmt.Errorf("Expected allow on all other to external endpoints but got deny")
+	out, err := fw.PodManager.ExecInPod(namespace, podName, cmd)
+	if err != nil {
+		GinkgoWriter.Printf("tcpProbe %s:%d exec error: %v\n", host, port, err)
 	}
-	return nil
+	return strings.TrimSpace(out)
```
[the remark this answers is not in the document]
yash97: Fixed in b22e0c1. tcpProbe now returns a `PROBE_ERROR` sentinel on ExecInPod failure instead of an empty string, so an exec (infra) failure is explicit in the Eventually/Consistently output rather than showing as `""`. Behavior is intentionally unchanged — `PROBE_ERROR`, like the old empty string, is neither `OPEN` nor `CLOSE`, so Eventually still retries past a transient blip; this is a diagnostics improvement, not a change to any assertion outcome.

test/integration/policy/except_block_test.go
```diff
@@ -172,49 +167,65 @@ var _ = Describe("IPBlock Except Test Cases", func() {
...
 		It("should allow on server prefix and 3306 port, deny on rest server-prefix ports, allow all on rest of endpoints", func() {
-			time.Sleep(30 * time.Second)
+			cfg := ipFamilyConfigForIP(serverIP)
 
-			// fetch the logs
-			logs, err := fw.PodManager.PodLogs(clientNamespace, clientName)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(processIPBlockLogs(logs, allowPort, blockPort)).To(Succeed())
+			// Deny converging first proves enforcement is active.
+			By(fmt.Sprintf("denying egress to the server on excepted port %d", blockPort), func() {
```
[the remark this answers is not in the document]
yash97: Updated the PR description. The "Fix" section now describes the exec-based `Eventually`/`Consistently` probing (via `tcpProbe`) with the fixed sleeps and log parsing removed, matching the current implementation; also dropped the stale "happy to switch to a poll/Eventually" note since that is now the approach.

viveksb007 (approved):

Issue #613, opened 2026-07-10
Memory leak: podIdentifierToPolicyEndpointMap inserts entries for all pods cluster-wide but only removes node-local ones


mikhailshilkov:
**What happened**:

On a ~1,000-node cluster with high pod churn (~150 pod creates+deletes/s, uniquely-named pods) under 3 long-lived egress NetworkPolicies (~45 PolicyEndpoint slices, ~35k selected pods), every aws-eks-nodeagent leaks Go heap at ~420 MiB/h until OOM-killed (~every 10h at a 4Gi limit), fleet-wide. Each restart causes a window of `plugin type="aws-cni" failed (add): add cmd: failed to setup network policy` pod-create failures.

Growth is identical on nodes hosting zero policy-selected pods. go_memstats_heap_inuse_bytes tracks container age linearly (608Mi @ 1.9h → 1342Mi @ 4.5h → 1570Mi @ 4.8h; RSS ≈ heap). A read-only scan of one agent's live heap (800MiB sample via /proc/<pid>/mem) found ~600k copies of our pod-name prefix (~17× the number of live pods) and ~54k copies of a single NetworkPolicy's name (retained parentPEList slices) i.e. per-pod bookkeeping for pods that no longer exist.

Note that pods must be uniquely-named and names remain unique after GetPodIdentifier strips the final dash-segment, e.g. `job-sb-019f3e0d-441f-…-<rand>` → identifier `job-sb-019f3e0d-441f-…` (one per pod).

Root cause appears visible in v1.3.5 controllers/policyendpoints_controller.go:

- Insertion is cluster-wide: deriveTargetPods calls updatePodIdentifierToPEMap(podIdentifier, parentPEList) at [L622](https://github.com/aws/aws-network-policy-agent/blob/v1.3.5/controllers/policyendpoints_controller.go#L622), outside the nodeIP.Equal(pod.HostIP) check at L617 — so every agent stores an entry for every selected pod in the cluster.
- Removal is local-only: targetPodIdentifiers is populated only inside the node-IP match (L617–621); [deriveStalePodIdentifiers (L680)](https://github.com/aws/aws-network-policy-agent/blob/v1.3.5/controllers/policyendpoints_controller.go#L680-L700) computes staleness against that node-local set only.
- `cleanUpPolicyEndpoint` runs only on PE deletion, which never happens while the parent NetworkPolicy exists (pod deletion = slice update).

[...]

PR #615, opened 2026-07-13
fix: clean up non-local pod identifiers from the PolicyEndpoint and ClusterPolicyEndpoint pod identifier maps

SokolAndrey:
updatePodIdentifierToPEMap registers every pod selected by a PolicyEndpoint, including pods on other nodes, so the gRPC handler can apply policies at pod launch. However, stale entry cleanup only covered pod identifiers local to the node: networkPolicyToPodIdentifierMap was populated from local pods only, and the PE delete path only purged local podsToBeCleanedUp. As a result, every remote pod identifier ever selected by a policy stayed in the map forever, and pod churn from rollouts, jobs and cronjobs grew it without bound (~66MB retained in 25 minutes on a busy cluster, per heap profile diff).

Non-local entries are still registered, preserving at-launch policy application; they are just cleanable now.

Attached 2 profiles taken in a span of 2 hours on a cluster with high pod churn.
[heap-t0.pb.gz](https://github.com/user-attachments/files/29969299/heap-t0.pb.gz)
[heap-t3.pb.gz](https://github.com/user-attachments/files/29969304/heap-t3.pb.gz)

Also added unit tests for both controllers; they fail without the fix and pass with it.

*Issue #, if available:*
#613 

*Description of changes:*
This fixes the same leak in both the PolicyEndpoints controller (`podIdentifierToPolicyEndpointMap`) and the ClusterPolicyEndpoints controller (`podIdentifierToClusterPolicyEndpointMap`), which register pod identifiers the same way.

### DIFF
--- a/controllers/clusterpolicyendpoints_controller.go
+++ b/controllers/clusterpolicyendpoints_controller.go
@@ -348,6 +348,7 @@ func (r *ClusterPolicyEndpointsReconciler) deriveTargetPodsForParentCNP(ctx cont
 	var newTargetPods, podsToBeCleanedUp, currentPods []npatypes.Pod
 	var targetPodIdentifiers []string
 	podIdentifiers := make(map[string]bool)
+	allPodIdentifiers := make(map[string]bool)
 
 	// Get current pods selected by the PE objects for this CNP
 	existingPods, podsPresent := r.ClusterPolicyEndpointSelectorMap.Load(resourceName)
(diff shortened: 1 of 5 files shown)


jaydeokar:
Thanks  @SokolAndrey for making the change. The approach looks good, I'm running some tests on my end. 
We have to make a similar change in the clusterpolicyendpoint controller too.

SokolAndrey:
> Thanks @SokolAndrey for making the change. The approach looks good, I'm running some tests on my end. We have to make a similar change in the clusterpolicyendpoint controller too.

Happy to help! ;) Added similar fix and a unit test for clusterpolicyendpoint controller.

jaydeokar:
There's a different fix that will be needed since we are now cleanly deleting all the left over PEs. https://github.com/aws/aws-network-policy-agent/issues/616

I'm testing the changes currently with that fix

PR #619, opened 2026-07-17
fix: validate loaded BPF program has associated maps before use

Pavani-Panakanti:
loadBPFProgram trusts that a successful LoadBpfFile always returns a program with its associated maps linked which will not stay true in error cases

Validate the loaded program before returning it:
- error if no BpfData exists at the expected pinPath
- error if the program FD is 0
- error if the program has no associated maps

Returning an error causes AttacheBPFProbes to fail so the controller retries the reconcile with a freshly loaded program ( instead of silently persisting broken state. 

related issue - https://github.com/aws/aws-ebpf-sdk-go/pull/157

### DIFF
--- a/pkg/ebpf/bpf_client.go
+++ b/pkg/ebpf/bpf_client.go
@@ -976,10 +976,27 @@ func (l *bpfClient) loadBPFProgram(fileName string, direction string,
 		return nil, -1, err
 	}
 
+	// Validate the loaded program before returning it. A successful LoadBpfFile
+	// must yield a program with a valid FD and its associated maps linked.
 	pinPath := utils.GetBPFPinPathFromPodIdentifier(podIdentifier, direction)
-	progFD := progInfo[pinPath].Program.ProgFD
+	bpfData, ok := progInfo[pinPath]
+	if !ok {
+		sdkAPIErr.WithLabelValues("LoadBpfFile-NoPinPath").Inc()
+		return nil, -1, fmt.Errorf("no program data found at pinPath %s for pod %s direction %s", pinPath, podIdentifier, direction)
+	}
 
-	log().Infof("Prog Load Succeeded for %s, progFD: %d, pinpath: %s", direction, progFD, pinPath)
+	progFD := bpfData.Program.ProgFD
+	if progFD == 0 {
+		sdkAPIErr.WithLabelValues("LoadBpfFile-InvalidFD").Inc()
+		return nil, -1, fmt.Errorf("program loaded with invalid FD 0 for pod %s direction %s", podIdentifier, direction)
+	}
+
+	if len(bpfData.Maps) == 0 {
+		sdkAPIErr.WithLabelValues("LoadBpfFile-NoMaps").Inc()
+		return nil, -1, fmt.Errorf("program loaded but has no associated maps for pod %s direction %s progFD %d", podIdentifier, direction, progFD)
+	}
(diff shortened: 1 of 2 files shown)


pkg/ebpf/bpf_client.go
```diff
@@ -976,10 +976,27 @@ func (l *bpfClient) loadBPFProgram(fileName string, direction string,
...
+	progFD := bpfData.Program.ProgFD
+	if progFD == 0 {
+		sdkAPIErr.WithLabelValues("LoadBpfFile").Inc()
+		return nil, -1, fmt.Errorf("program loaded with invalid FD 0 for pod %s direction %s", podIdentifier, direction)
+	}
+
+	if len(bpfData.Maps) == 0 {
+		sdkAPIErr.WithLabelValues("LoadBpfFile").Inc()
+		return nil, -1, fmt.Errorf("program loaded but has no associated maps for pod %s direction %s progFD %d", podIdentifier, direction, progFD)
+	}
```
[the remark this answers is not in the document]
Pavani-Panakanti: issue is still coming from load program

pkg/ebpf/bpf_client.go
```diff
@@ -976,10 +976,27 @@ func (l *bpfClient) loadBPFProgram(fileName string, direction string,
 		return nil, -1, err
 	}
 
+	// Validate the loaded program before returning it. A successful LoadBpfFile
+	// must yield a program with a valid FD and its associated maps linked.
 	pinPath := utils.GetBPFPinPathFromPodIdentifier(podIdentifier, direction)
-	progFD := progInfo[pinPath].Program.ProgFD
+	bpfData, ok := progInfo[pinPath]
+	if !ok {
+		sdkAPIErr.WithLabelValues("LoadBpfFile").Inc()
```
jaydeokar: Possible to add more dimensions to this based on what fails ?  Like adding an error message as a label value? 
By this even if logs get rotated, we'll have some datapoints on what the failure was

jaydeokar (approved):
lgtm

pkg/ebpf/bpf_client.go
viveksb007, line 973: shouldn't this err handle the case where pinPath doesn't exist in progInfo?

is it a valid case where sdk will not throw err and also give an invalid progInfo?

pkg/ebpf/bpf_client.go
```diff
@@ -976,10 +976,27 @@ func (l *bpfClient) loadBPFProgram(fileName string, direction string,
...
+	progFD := bpfData.Program.ProgFD
+	if progFD == 0 {
+		sdkAPIErr.WithLabelValues("LoadBpfFile-InvalidFD").Inc()
+		return nil, -1, fmt.Errorf("program loaded with invalid FD 0 for pod %s direction %s", podIdentifier, direction)
+	}
+
+	if len(bpfData.Maps) == 0 {
+		sdkAPIErr.WithLabelValues("LoadBpfFile-NoMaps").Inc()
+		return nil, -1, fmt.Errorf("program loaded but has no associated maps for pod %s direction %s progFD %d", podIdentifier, direction, progFD)
+	}
```
viveksb007, line 997: it seems we are checking integrity of `bpfData` here, we can move it out in common method, i think we do  this check at few places.

since we are checking the ProgFD, can there be cases where we need to check MapFD too?

PR #622, opened 2026-07-21
conntrack: summarize stale-entry cleanup logging to cut log noise

yash97:
## What

The userspace conntrack GC logs one `info` line per stale entry it deletes, every cleanup cycle (`CleanupConntrackMap` / `Cleanupv6ConntrackMap`). On an idle cluster this single line dominates the agent's log output — in one measurement it was ~84–98% of all lines (hundreds of lines every 5‑minute cycle with essentially no workload).

## Change

- Demote the per-entry `Conntrack cleanup Delete - Conntrack Key : ...` line from `info` to `debug` (full 5-tuple detail preserved).
- Emit a single per-cycle `info` summary instead: `Done cleanup of conntrack map: deleted N stale entries out of M tracked (per-entry detail at debug level)`.
- Applied symmetrically to both the IPv4 and IPv6 cleanup paths.

## Why this shape

The shipped default log level is `debug`, so this is intentionally **not** a silent drop:
- At `--log-level=debug` (default): behavior is unchanged in content — every deleted flow's full tuple is still logged, just correctly tagged `debug`, plus the new summary. This preserves debuggability for flow-level issues (e.g. the port-reuse GC race), where the source port of the evicted flow is exactly what you need.
- At `--log-level=info`: the per-entry flood disappears and you get one summary line per cycle with the deleted/tracked counts. This makes `info` a usable, low-volume mode — previously the flood *was* info, so there was no way to shed it without losing everything.

## Live A/B test (fresh nodes, identical load)

Built into an image and deployed on a 3-node EKS cluster. Nodes were **recycled via ASG instance refresh first** so each started with an empty conntrack map and a fresh log file (rules out any pre-existing entries). Then the same network-policy'd churn workload (client pods doing tight HTTP request loops → large short-lived flow / DNS turnover) was run for two back-to-back windows on one node, changing **only** the log level between them:

| | DEBUG window (10 min) | INFO window (~11 min) |
|---|---|---|
| GC delete cycles observed | 3 | 3 |
| Stale entries GC deleted (sum over window) | 51,193 | 181,107 |
| **Per-entry `Conntrack cleanup Delete` lines emitted** | **51,193** (all `level:debug`) | **0** |
| Per-cycle summary lines emitted | 3 | 3 |
| **Total new log lines in window** | **~51,555** | **10** |

Same GC work in both windows (tens of thousands of stale entries deleted per cycle); the only difference is the log level. At `info`, the per-entry flood is fully suppressed and each cycle collapses to a single summary line — the three info-window cycles deleted 75,313 / 51,433 / 54,361 entries respectively yet produced just **10 total log lines**.

**INFO — per-cycle summary (verbatim):**
```json
{"level":"info","msg":"Done cleanup of conntrack map: deleted 75313 stale entries out of 80564 tracked (per-entry detail at debug level)"}
```

**DEBUG — per-entry line, full detail preserved (verbatim):**
```json
{"level":"debug","msg":"Conntrack cleanup Delete - Conntrack Key : Source IP - 192.168.27.168 Source port - 55006 Dest IP - 10.100.0.10 Dest port - 53 Protocol - 17 Owner IP - 192.168.27.168 Ifindex - 828"}
```

**BEFORE — same line on current `main` (emitted at `info`, one per deleted entry):**
```json
{"level":"info","msg":"Conntrack cleanup Delete - Conntrack Key : Source IP - 192.168.27.168 Source port - 55006 Dest IP - 10.100.0.10 Dest port - 53 Protocol - 17 Owner IP - 192.168.27.168 Ifindex - 828"}
```

> Note: the per-cycle counts above are from a synthetic high-churn workload chosen to make the effect measurable; they are not steady-state. Near-idle, cleanup deletes ~180–320 entries/cycle — which today is still ~180–320 `info` lines/cycle and becomes 1 summary line with this change.

## Testing

- `go build ./...`, `go vet ./pkg/ebpf/conntrack/...` clean.
- `go test ./pkg/ebpf/... ./controllers/... ./pkg/rpc/... ./pkg/fwruleprocessor/...` pass (no test asserts on these log lines).
- Live fresh-node A/B validation as tabulated above: at `info` the conntrack-delete flood collapses to one summary per cycle; at `debug` full per-entry detail remains.


pkg/ebpf/conntrack/conntrack_client.go
```diff
@@ -190,14 +190,14 @@ func (c *conntrackClient) CleanupConntrackMap() {
...
+				// Per-entry detail at debug; the info summary below reports the count.
+				log().Debugf("Conntrack cleanup Delete - Conntrack Key : Source IP - %s Source port - %d Dest IP - %s Dest port - %d Protocol - %d Owner IP - %s Ifindex - %d", utils.ConvIntToIPv4(expiredFlow.Source_ip).String(), expiredFlow.Source_port, utils.ConvIntToIPv4(expiredFlow.Dest_ip).String(), expiredFlow.Dest_port, expiredFlow.Protocol, utils.ConvIntToIPv4(expiredFlow.Owner_ip).String(), expiredFlow.Ifindex)
 				c.conntrackMap.DeleteMapEntry(uintptr(unsafe.Pointer(&expiredFlow)))
-
+				deletedEntries++
 			}
 		}
 		//c.localConntrackV4Cache = make(map[utils.ConntrackKey]bool)
-		log().Info("Done cleanup of conntrack map")
+		log().Infof("Done cleanup of conntrack map: deleted %d stale entries out of %d tracked (per-entry detail at debug level)", deletedEntries, len(c.localConntrackV4Cache))
```
Pavani-Panakanti: this gets printed every time we can remove it. Its a standard pattern and not that much value adding it - (per-entry detail at debug level). It can be confusing as its not related to conntrack cleanup info

pkg/ebpf/conntrack/conntrack_client.go
```diff
@@ -343,22 +343,25 @@ func (c *conntrackClient) Cleanupv6ConntrackMap() {
 
 		}
 		// Check if the local cache and kernel cache is in sync
+		deletedEntries := 0
+		trackedEntries := len(c.localConntrackV6Cache)
```
Pavani-Panakanti, line 347: can we do same for ipv4 as well - for readability and less error prone to any changes we might make in future to change that cache values similar to ipv6

Pavani-Panakanti (approved):

PR #623, opened 2026-07-21
perf: add trie for cidr lookup

srikanthjg:
## Summary

When many pods share a single NetworkPolicy, the linear scan dominates reconcile time and allocation churn. Profiling showed `checkAndDeriveL4InfoFromAnyMatchingCIDRs` plus `net.ParseCIDR` accounting for the majority of CPU, and the resulting allocation rate could grow memory unbounded and OOM the agent at high rule counts. Replacing the O(rules^2) scan with the trie removes that hot path.

## Fix

Replace the linear CIDR scan in `checkAndDeriveL4InfoFromAnyMatchingCIDRs` with a binary prefix trie.

Previously, for every rule the code compared the address against every previously-seen CIDR and re-parsed each CIDR with `net.ParseCIDR` — O(rules) per rule, O(rules^2) overall. The trie walks the address bits once to find all containing prefixes, reducing the containment lookup to O(address-bits): 32 for IPv4, 128 for IPv6, independent of how many CIDRs are present.

The trie keeps separate per-family (IPv4 / IPv6) sub-trees behind a single `normalizeIP` entry point, so both address families share one code path. The v4-mapped IPv6 case (`::ffff:x.x.x.x` prefixes) is normalized so mapped CIDRs are matched against the correct family instead of being silently dropped.

## Wins
- CPU:  Per-rule CIDR containment check(checkAndDeriveL4InfoFromAnyMatchingCIDRs) — drops from 73% → 4% cumulative CPU at scale(15 pods × 15PE x 200 rules)
- Latency: Reconcile p99 improves 4–10x depending on scale.
- Allocations: 26% fewer objects per reconcile cycle; key string-splitting allocations eliminated.
- Memory: No steady-state increase — trie nodes are transient, no leak risk.
- Complexity: The trie replaces O(rules) per-address containment checks with O(32) for IPv4 / O(128) for IPv6, independent of rule count. This reduces overall complexity of reconcilation from O(pods)*O(rules^2) to O(pods)xO(rules * 32)

## Testing

Correctness was validated on both IPv4 and IPv6 clusters.

| Test | IPv4 | IPv6 |
|---|---|---|
| Unit tests | Pass | Pass |
| Network Policy integration suite | 6 / 6 | 6 / 6 |
| eBPF conntrack integration suite | 3 / 3 | n/a (suite is IPv4-only) |
| Cyclonus network-policy conformance | 112 / 112 | 112 / 112 |

Regarding this test file change `test/integration/policy/except_block_test.go`: 
This test was flaky and it was failing randomly. This is a test-only change (no production code), it fixes a pre-existing flaky tests that fails on baseline too (not something the trie introduced). 

## Performance

Measured on a single node with one NetworkPolicy shared across N pods, 50 rules per pod. Reconcile latency is taken from the agent's built-in reconcile-time metric (p99). "eBPF" is average per-call eBPF map-update latency. number of pe is equal to the number of pods. 

| Pods | p99 reconcile before | p99 reconcile after | eBPF before | eBPF after |
|---|---|---|---|---|
| 15 | 685 ms | 146 ms | 23.2 ms | 4.4 ms |
| 30 | 4,971 ms | 570 ms | 91.2 ms | 9.4 ms |
| 80 | 60,000 ms | 6,962 ms (queue drains) | 492 ms | 49 ms |

## More details

Config for pprof shared below: 80pods,80PEs,50rules per PE

### CPU Pprof data
Baseline:
```
         0     0%     0%      9.89s 79.76%  ...fwruleprocessor.(*FirewallRuleProcessor).ComputeMapEntriesFromEndpointRules
     0.35s  2.82%  2.82%      9.08s 73.23%  ...fwruleprocessor.checkAndDeriveL4InfoFromAnyMatchingCIDRs
     0.57s  4.60%  7.42%      7.87s 63.47%  net.ParseCIDR
     0.73s  5.89% 55.32%      0.73s  5.89%  internal/runtime/syscall/linux.Syscall6
     0.01s 0.081% 55.40%      0.73s  5.89%  syscall.RawSyscall6
         0     0% 59.68%      0.58s  4.68%  ...logger.(*structuredLogger).Infof
         0     0% 59.68%      0.58s  4.68%  go.uber.org/zap.(*SugaredLogger).Infof (inline)
     0.01s 0.081% 87.34%      0.10s  0.81%  ...fwruleprocessor.mergeDuplicateL4Info
```

WIth fix:
```
     0.37s  0.61%  0.61%     44.59s 73.19%  github.com/aws/aws-network-policy-agent/pkg/fwruleprocessor.(*FirewallRuleProcessor).ComputeMapEntriesFromEndpointRules
     0.07s  0.11%  0.89%     17.82s 29.25%  github.com/aws/aws-network-policy-agent/pkg/logger.(*structuredLogger).Infof
     0.83s  1.36%  2.25%     15.39s 25.26%  github.com/aws/aws-network-policy-agent/pkg/fwruleprocessor.mergeDuplicateL4Info
     0.10s  0.16%  2.58%     11.48s 18.84%  syscall.RawSyscall6
    11.38s 18.68% 21.26%     11.38s 18.68%  internal/runtime/syscall/linux.Syscall6
     0.17s  0.28% 38.33%      2.60s  4.27%  net.ParseCIDR
     0.15s  0.25% 38.72%      2.40s  3.94%  github.com/aws/aws-network-policy-agent/pkg/fwruleprocessor.checkAndDeriveL4InfoFromAnyMatchingCIDRsTrie
     0.36s  0.59% 77.28%      0.50s  0.82%  github.com/aws/aws-network-policy-agent/pkg/fwruleprocessor.(*cidrTrie).findContainingKeys
```

Comparision:
**Side-by-side comparison (same config):**

| Function | Baseline (cum %) | Trie (cum %) | Change |
|---|---|---|---|
| `ComputeMapEntriesFromEndpointRules` | **79.76%** | **73.19%** | (parent -- similar, but contents shifted) |
| `checkAndDeriveL4InfoFromAnyMatchingCIDRs` | **73.23%** | -- | eliminated |
| `checkAndDeriveL4InfoFromAnyMatchingCIDRsTrie` | -- | **3.94%** | replacement |
| `net.ParseCIDR` | **63.47%** | **4.27%** | -93% (parse once at trie build) |
| `Syscall6` (eBPF map updates) | 5.89% | 18.68% | same absolute work, higher % share |
| `syscall.RawSyscall6` | 5.89% | 18.84% | same absolute work, higher % share |
| `Infof` (logging) | 4.68% | 29.25% | same absolute work, higher % share |
| `mergeDuplicateL4Info` | 0.81% | 25.26% | now visible (was hidden behind scan) |
| `cidrTrie.findContainingKeys` | -- | 0.82% | new (O(prefix-bits) lookup) |
| `cidrTrie.insert` | -- | 0.67% | new (trie construction) |
| `ComputeTrieValue` | -- | 2.81% | new (value computation) |

The CIDR-matching share of CPU drops from 73% to under 4% at the identical config. With the scan gone, reconcile CPU is now dominated by logging (29.3%) and eBPF map update syscalls (18.7%) -- I/O, not rule computation.

### Memory Pprof data

Both filtered with: `grep -Ei "genSplit|Sprintf|Infof|ParseCIDR|checkAndDerive|trie|ComputeMap|mergeDuplicate|LoadProg"`

Baseline:
```
Total alloc_objects = 533,572

     65536 12.28%  ...  65536 12.28%  strings.genSplit
     43691  8.19%  ... 43691  8.19%  fmt.Sprintf
         0     0%  ... 36503  6.84%  go.uber.org/zap.(*SugaredLogger).Infof (inline)
         0     0%  ... 36275  6.80%  ...ebpf-sdk-go/pkg/logger.(*structuredLogger).Infof
```

With fix:
```
Total alloc_objects = 395,511

         0     0%  ... 11261  2.85%  go.uber.org/zap.(*SugaredLogger).Infof (inline)
     10923  2.76%  ... 10974  2.77%  ...elfparser.(*elfLoader).loadProg
         0     0%  ...  6555  1.66%  ...ebpf-sdk-go/pkg/logger.(*structuredLogger).Infof
      3277  0.83%  ...  5798  1.47%  fmt.Sprintf
         0     0%  ...  4706  1.19%  ...pkg/logger.(*structuredLogger).Infof
```

**Allocation comparison:**

| Function | Baseline (objects) | Trie (objects) | Change |
|---|---|---|---|
| Total allocations | **533,572** | **395,511** | **-26%** |
| `strings.genSplit` | 65,536 (12.28%) | 0 | **eliminated** |
| `fmt.Sprintf` | 43,691 (8.19%) | 5,798 (1.47%) | **-87%** |
| `Infof` (logging) | 36,503 (6.84%) | 11,261 (2.85%) | proportional to reconcile count |

The trie eliminates repeated CIDR string splitting (`strings.genSplit` -- the linear scan re-parsed every CIDR on every comparison) and most `fmt.Sprintf` calls (format strings for error paths that no longer execute).

**Heap Comparision**:
No significant changes. The trie nodes (`ComputeTrieValue` -- 10.5MB) are the primary heap resident. These are built once per reconcile and freed after map updates complete. The baseline heap at 80px50r was 27.3MB vs trie 26.8MB (effectively equal -- dominated by the eBPF processing not rule computation).

### DIFF
--- a/pkg/fwruleprocessor/cidr_trie.go
+++ b/pkg/fwruleprocessor/cidr_trie.go
@@ -1,0 +1,123 @@
+package fwruleprocessor
+
+import (
+	"net"
+)
+
+type cidrTrieNode struct {
+	children [2]*cidrTrieNode
+	cidrKeys []string
+}
+
+// cidrTrie keeps IPv4 and IPv6 in separate sub-tries to avoid cross-family prefix collisions.
+type cidrTrie struct {
+	v4Root *cidrTrieNode
+	v6Root *cidrTrieNode
+}
+
+func newCIDRTrie() *cidrTrie {
+	return &cidrTrie{
+		v4Root: &cidrTrieNode{},
+		v6Root: &cidrTrieNode{},
+	}
+}
+
+// normalizeIP returns the IP in its canonical form (4 bytes for IPv4, 16 for IPv6)
+// and whether it belongs to the v6 family. v4-mapped IPv6 addresses (::ffff:x.x.x.x)
+// are normalized to their 4-byte IPv4 form so they match against the v4 sub-trie.
+func normalizeIP(ip net.IP) (normalized net.IP, isV6 bool) {
+	if v4 := ip.To4(); v4 != nil {
+		return v4, false
+	}
+	if v6 := ip.To16(); v6 != nil {
+		return v6, true
+	}
+	return nil, false
+}
+
+func (t *cidrTrie) rootFor(isV6 bool) *cidrTrieNode {
+	if isV6 {
+		return t.v6Root
+	}
+	return t.v4Root
+}
+
+// insert adds a CIDR key to the trie at the position determined by the network
+// prefix bits. The key is stored at the node corresponding to the last prefix bit,
+// so findContainingKeys can collect it when traversing an address that falls within
+// this prefix.
+func (t *cidrTrie) insert(cidrStr string, ipNet *net.IPNet) {
+	if ipNet == nil {
+		log().Debugf("cidrTrie.insert: skipping key %q: nil ipNet", cidrStr)
+		return
+	}
+	ip, isV6 := normalizeIP(ipNet.IP)
+	if ip == nil {
+		log().Debugf("cidrTrie.insert: skipping key %q: unparseable network IP", cidrStr)
+		return
+	}
+
+	ones, bits := ipNet.Mask.Size()
+	if bits == 0 {
+		log().Debugf("cidrTrie.insert: skipping key %q: non-canonical mask", cidrStr)
+		return
+	}
+
+	if !isV6 && bits == 128 {
+		ones -= 96
+	}
+
+	maxBits := len(ip) * 8
+	if ones < 0 || ones > maxBits {
+		log().Debugf("cidrTrie.insert: skipping key %q: prefix length %d out of range for %d-bit family", cidrStr, ones, maxBits)
+		return
+	}
+
+	node := t.rootFor(isV6)
+	for i := 0; i < ones; i++ {
+		byteIdx := i / 8
+		bitIdx := 7 - (i % 8)
+		bit := (ip[byteIdx] >> uint(bitIdx)) & 1
+
+		if node.children[bit] == nil {
+			node.children[bit] = &cidrTrieNode{}
+		}
+		node = node.children[bit]
+	}
+	node.cidrKeys = append(node.cidrKeys, cidrStr)
+}
+
+// findContainingKeys walks the trie from root to leaf along the bits of ip,
+// collecting all CIDR keys whose prefix contains the given address.
+// Returns results from shortest prefix (closest to root) to longest.
+func (t *cidrTrie) findContainingKeys(ip net.IP) []string {
+	var result []string
+
+	ipBytes, isV6 := normalizeIP(ip)
+	if ipBytes == nil {
+		return result
+	}
+
+	node := t.rootFor(isV6)
+	if len(node.cidrKeys) > 0 {
+		result = append(result, node.cidrKeys...)
+	}
+
+	maxBits := len(ipBytes) * 8
+	for i := 0; i < maxBits; i++ {
+		byteIdx := i / 8
+		bitIdx := 7 - (i % 8)
+		bit := (ipBytes[byteIdx] >> uint(bitIdx)) & 1
+
+		if node.children[bit] == nil {
+			break
+		}
+		node = node.children[bit]
+
+		if len(node.cidrKeys) > 0 {
+			result = append(result, node.cidrKeys...)
+		}
+	}
+
+	return result
+}
--- a/pkg/fwruleprocessor/cidr_trie.go
+++ b/pkg/fwruleprocessor/cidr_trie.go
@@ -46,9 +46,10 @@ func (t *cidrTrie) rootFor(isV6 bool) *cidrTrieNode {
 // prefix bits. The key is stored at the node corresponding to the last prefix bit,
 // so findContainingKeys can collect it when traversing an address that falls within
 // this prefix.
-func (t *cidrTrie) insert(cidrStr string, ipNet *net.IPNet) {
-	if ipNet == nil {
-		log().Debugf("cidrTrie.insert: skipping key %q: nil ipNet", cidrStr)
+func (t *cidrTrie) insert(cidrStr string) {
+	_, ipNet, err := net.ParseCIDR(cidrStr)
+	if err != nil || ipNet == nil {
+		log().Debugf("cidrTrie.insert: skipping key %q: parse error: %v", cidrStr, err)
 		return
 	}
 	ip, isV6 := normalizeIP(ipNet.IP)
--- a/pkg/fwruleprocessor/cidr_trie.go
+++ b/pkg/fwruleprocessor/cidr_trie.go
@@ -6,7 +6,7 @@ import (
 
 type cidrTrieNode struct {
 	children [2]*cidrTrieNode
-	cidrKeys []string
+	cidrKey  string
 }
 
 // cidrTrie keeps IPv4 and IPv6 in separate sub-tries to avoid cross-family prefix collisions.
@@ -65,6 +65,9 @@ func (t *cidrTrie) insert(cidrStr string) {
 	}
 
 	if !isV6 && bits == 128 {
+		// v4-mapped IPv6 CIDR (e.g. ::ffff:10.0.0.0/104): normalizeIP converted
+		// the IP to 4-byte v4 form, but the mask still reports 128-bit width.
+		// Subtract the 96-bit v6 prefix to get the effective v4 prefix length.
 		ones -= 96
 	}
 
@@ -85,7 +88,9 @@ func (t *cidrTrie) insert(cidrStr string) {
 		}
 		node = node.children[bit]
 	}
-	node.cidrKeys = append(node.cidrKeys, cidrStr)
+	// Store the canonical form (host bits masked) so the key matches
+	// what the caller uses in the nonHostCIDRs map.
+	node.cidrKey = ipNet.String()
 }
 
 // findContainingKeys walks the trie from root to leaf along the bits of ip,
@@ -100,8 +105,8 @@ func (t *cidrTrie) findContainingKeys(ip net.IP) []string {
 	}
 
 	node := t.rootFor(isV6)
-	if len(node.cidrKeys) > 0 {
-		result = append(result, node.cidrKeys...)
+	if node.cidrKey != "" {
+		result = append(result, node.cidrKey)
 	}
 
 	maxBits := len(ipBytes) * 8
@@ -115,8 +120,8 @@ func (t *cidrTrie) findContainingKeys(ip net.IP) []string {
 		}
 		node = node.children[bit]
 
-		if len(node.cidrKeys) > 0 {
-			result = append(result, node.cidrKeys...)
+		if node.cidrKey != "" {
+			result = append(result, node.cidrKey)
 		}
 	}
 
--- a/pkg/fwruleprocessor/cidr_trie_test.go
+++ b/pkg/fwruleprocessor/cidr_trie_test.go
@@ -166,6 +166,32 @@ func buildTrie(ruleMap map[string]EbpfFirewallRules) *cidrTrie {
 	return trie
 }
 
+// checkAndDeriveL4InfoFromAnyMatchingCIDRs is the original O(N) linear scan.
+// Retained here as the oracle for TestTrieVsLinear_Randomized.
+func checkAndDeriveL4InfoFromAnyMatchingCIDRs(firewallRule string,
+	cidrsMap map[string]EbpfFirewallRules) []v1alpha1.Port {
+	var matchingCIDRL4Info []v1alpha1.Port
+
+	_, ipToCheck, _ := net.ParseCIDR(firewallRule)
+	for cidr, cidrFirewallInfo := range cidrsMap {
+		_, cidrEntry, _ := net.ParseCIDR(cidr)
+		if cidrEntry.Contains(ipToCheck.IP) {
+			foundInExcept := false
+			for _, except := range cidrFirewallInfo.Except {
+				_, exceptEntry, _ := net.ParseCIDR(string(except))
+				if exceptEntry.Contains(ipToCheck.IP) {
+					foundInExcept = true
+					break
+				}
+			}
+			if !foundInExcept {
+				matchingCIDRL4Info = append(matchingCIDRL4Info, cidrFirewallInfo.L4Info...)
+			}
+		}
+	}
+	return matchingCIDRL4Info
+}
+
 func assertEquivalent(t *testing.T, ruleMap map[string]EbpfFirewallRules, target string) {
 	t.Helper()
 	linear := checkAndDeriveL4InfoFromAnyMatchingCIDRs(target, ruleMap)
@@ -271,7 +297,3 @@ func TestTrieVsLinear_Randomized(t *testing.T) {
 		}
 	}
 }
-
-// --- Concurrency safety ---
-
-
--- a/pkg/fwruleprocessor/cidr_trie_test.go
+++ b/pkg/fwruleprocessor/cidr_trie_test.go
@@ -19,17 +19,7 @@ func insertAll(t *testing.T, trie *cidrTrie, cidrs ...string) {
 	}
 }
 
-func TestCIDRTrie_Insert_NilIPNet(t *testing.T) {
-	trie := newCIDRTrie()
-	trie.insert("not-a-cidr")
-	assert.Empty(t, trie.findContainingKeys(net.ParseIP("10.1.2.3")))
-}
-
-func TestCIDRTrie_FindContainingKeys_NilIP(t *testing.T) {
-	trie := newCIDRTrie()
-	insertAll(t, trie, "10.0.0.0/8")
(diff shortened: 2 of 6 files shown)


viveksb007 (commented):
can you add flame graphs and pprof which shows time spent on fns?

that should also show diff b/w time diff in Compute* method.

pkg/fwruleprocessor/fw_rule_processor.go
viveksb007: since this fn `checkAndDeriveL4InfoFromAnyMatchingCIDRs` is used as oracle to compare new impl, we need to remove it at some point. maybe add a TODO comment to clean it up

pkg/fwruleprocessor/cidr_trie.go
```diff
@@ -0,0 +1,116 @@
...
+	if len(node.cidrKeys) > 0 {
+		result = append(result, node.cidrKeys...)
+	}
+
+	maxBits := len(ipBytes) * 8
+	for i := 0; i < maxBits; i++ {
+		byteIdx := i / 8
+		if byteIdx >= len(ipBytes) {
+			break
+		}
```
viveksb007: will this `if` condition ever execute?

since `maxBits := len(ipBytes) * 8` and we are iterating over it

pkg/fwruleprocessor/cidr_trie_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+func randV4IP(r *rand.Rand) string {
+	return fmt.Sprintf("%d.%d.%d.%d/32", r.Intn(256), r.Intn(256), r.Intn(256), r.Intn(256))
+}
+
+func randV6IP(r *rand.Rand) string {
+	return fmt.Sprintf("%x:%x:%x:%x::1/128", r.Intn(0x10000), r.Intn(0x10000), r.Intn(0x10000), r.Intn(0x10000))
+}
+
+func TestTrieVsLinear_Randomized(t *testing.T) {
```
viveksb007, line 252: Nice test. 

[nit] can we enhance it? 

`
Two gaps in the randomized generator, though: it never produces Except blocks or v4-mapped (::ffff:) CIDRs — those are only exercised by the curated cases. Adding random excepts (e.g. carve a sub-CIDR out of ~30% of generated rules) would extend the fuzz coverage to the except-skip path, which is the trickiest part of the equivalence.
`

pkg/fwruleprocessor/cidr_trie.go
```diff
@@ -0,0 +1,116 @@
...
+		return
+	}
+
+	ones, bits := ipNet.Mask.Size()
+	if bits == 0 {
+		log().Debugf("cidrTrie.insert: skipping key %q: non-canonical mask", cidrStr)
+		return
+	}
+
+	if !isV6 && bits == 128 {
```
viveksb007, line 67: cidr trie isn't much intuitive to understand, can you add some simple doc strings for future readers to understand better on what/why we are doing it (we can avoid large explanations, just brief hints/explanation is fine for reader to get some idea)

viveksb007 (commented):
CR details also talk about `fmt.Sprintf`? how is that getting reduced with this trie introduction?

overall lgtm.

pkg/fwruleprocessor/fw_rule_processor.go
srikanthjg: >CR details also talk about fmt.Sprintf? how is that getting reduced with this trie introduction?

Trie builds CIDRs once at insert, eliminating repeated net.ParseCIDR (that in turn calls fmt.Sprintf); 

>maybe add a TODO comment to clean it up

sure, will do that.

pkg/fwruleprocessor/cidr_trie.go
srikanthjg: its a defensive check. we can remove it too.

pkg/fwruleprocessor/cidr_trie_test.go
srikanthjg: sure, i ll add random except generation.

pkg/fwruleprocessor/cidr_trie.go
srikanthjg: will do.

viveksb007 (commented):
thanks for adding the comments, its still difficult to grok the exact storage format of the IPs (v4 and v6) in the trie layout and we may need to add a doc.go to explain how IPs are being stored and retried in trie layout. 

I was checking and found few libs
- https://github.com/yl2chen/cidranger
- https://github.com/phemmer/go-iptrie

cidranger one seems promising, can you check once if its possible to re-use those?

srikanthjg:
> thanks for adding the comments, its still difficult to grok the exact storage format of the IPs (v4 and v6) in the trie layout and we may need to add a doc.go to explain how IPs are being stored and retried in trie layout.
> 
> I was checking and found few libs
> 
> * https://github.com/yl2chen/cidranger
> * https://github.com/phemmer/go-iptrie
> 
> cidranger one seems promising, can you check once if its possible to re-use those?

Checked both libraries. They don't fit our usage pattern.

Our trie isn't a generic IP in CIDR lookup. The hot path is: walk the trie for a given IP → collect every CIDR string key whose prefix contains it, use each key to look up the associated EbpfFirewallRules from a separate map which accumulate their L4 port info (skipping entries where the IP falls in an Except block). We need the original CIDR string back from the trie, not a net.IPNet or an interface.

cidranger — ContainingNetworks(ip) exists, but returns []RangerEntry (an interface wrapping net.IPNet). There's no way to get our CIDR string key back without wrapping every insert in a custom entry type — adds interface dispatch per node and the library itself uses the unmaintained dep tool (last commit 2019). No ::ffff: normalization either.

go-iptrie — Closer, but it normalizes all addresses to IPv6 internally (::-prefix everything), which conflicts with our ::ffff: → pure-v4 normalization. Returns []interface{} requiring type assertions on every collected result.

Our impl is ~100 lines, zero external dependencies, returns []string directly (the CIDR keys we need for the rules-map lookup), handles v4-mapped normalization in-line, and is validated by tests. For a transient per-reconcile structure holding 10-200 CIDRs, a full library adds dependency surface without simplifying the code.

Agree on the readability — will add a doc.go explaining the impl.

pkg/fwruleprocessor/cidr_trie.go
```diff
@@ -119,5 +119,9 @@ func (t *cidrTrie) findContainingKeys(ip net.IP) []string {
 		}
 	}
 
+	if len(result) > 0 {
```
viveksb007: this might result in a lot of debugging logs? should we move 1 level down to trace to something

viveksb007: IIRC, our default logging in MAO is debug, pls double check it once

viveksb007: default log level is debug https://github.com/aws/eks-charts/blob/5239609242f3430be50fb0b714a24d3ef2b47983/stable/aws-vpc-cni/values.yaml#L56

please double check that this doesn't create noisy logging at scale

pkg/fwruleprocessor/cidr_trie.go
```diff
@@ -0,0 +1,127 @@
...
+		return t.v6Root
+	}
+	return t.v4Root
+}
+
+// insert adds a CIDR key to the trie at the position determined by the network
+// prefix bits. The key is stored at the node corresponding to the last prefix bit,
+// so findContainingKeys can collect it when traversing an address that falls within
+// this prefix.
+func (t *cidrTrie) insert(cidrStr string, ipNet *net.IPNet) {
```
viveksb007: nit - can we simplify this fn signature as both params are related just different types

pkg/fwruleprocessor/cidr_trie.go
```diff
@@ -0,0 +1,127 @@
+package fwruleprocessor
+
+import (
+	"net"
+)
+
+type cidrTrieNode struct {
+	children [2]*cidrTrieNode
+	cidrKeys []string
```
viveksb007: can we remove this slice? i remember from offline discussion we can move it to `int` to signify the trie depth

srikanthjg: this is needed for cases for edge cases; for example when PE has different rules for 10.0.0.0/8 and also for 10.0.0.1/8 ; it is stored in the array as a separate entry; If this case will never happen on the controller, we can make it one entry.

pkg/fwruleprocessor/cidr_trie.go
srikanthjg: good call.

pkg/fwruleprocessor/cidr_trie.go
viveksb007: can you explain the case that this is handling in detail and its behavior impact?

for both `10.0.0.0/8` and `10.0.0.1/8`, only starting 8 bits are relevant for storage, so both will store ports are same level.

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -125,17 +126,18 @@ func (f *FirewallRuleProcessor) ComputeMapEntriesFromEndpointRules(firewallRules
...
-			// Check if the /m entry is part of any /n CIDRs that we've encountered so far
-			// If found, we need to include the port and protocol combination against the current entry as well since
-			// we use LPM TRIE map and the /m will always win out.
-			cidrL4Info = checkAndDeriveL4InfoFromAnyMatchingCIDRs(string(firewallRule.IPCidr), nonHostCIDRs)
+			cidrL4Info = checkAndDeriveL4InfoFromAnyMatchingCIDRsTrie(string(firewallRule.IPCidr), containmentTrie, nonHostCIDRs)
 			if len(cidrL4Info) > 0 {
 				firewallRule.L4Info = append(firewallRule.L4Info, cidrL4Info...)
 			}
 		}
 		cidrsMap[string(firewallRule.IPCidr)] = firewallRule
```
viveksb007, line 134: [unrelated to this PR change] need to double check the logic when same CIDR comes with different rules, do both survive or last write wins and we lose older rule

pkg/fwruleprocessor/cidr_trie.go
viveksb007: another thing to check if we are storing these 2 types as separate keys in kernel trie structure, does `10.0.0.0/8` and `10.0.0.1/8` results in 2 separate entries in kernel LPM trie

https://github.com/aws/aws-network-policy-agent/blob/main/pkg/utils/utils.go#L284

srikanthjg: this existed because of missing canonicalizing the IPs. We did have this bug(2 separate keys for such scenarios) but it was fixed only last month, from this commit -9fdad23 ; i have accomodated this in my latest rev.

pkg/fwruleprocessor/fw_rule_processor.go
srikanthjg: Before this commit 9fdad23 (fixed last month) last write wins. I have addressed canonicalization in the latest rev.

pkg/fwruleprocessor/cidr_trie_test.go
```diff
@@ -0,0 +1,275 @@
...
+				targets = append(targets, randV6IP(r))
+			}
+		}
+		for _, target := range targets {
+			assertEquivalent(t, ruleMap, target)
+		}
+	}
+}
+
+// --- Concurrency safety ---
```
viveksb007: is this some artifacts left?

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -214,6 +216,7 @@ func addDenyAllL4Entry(firewallRule *EbpfFirewallRules) {
 	firewallRule.L4Info = append(firewallRule.L4Info, denyAllL4Entry)
 }
 
+// TODO: Remove after trie validation is complete — retained only as the oracle for TestTrieVsLinear_Randomized.
```
viveksb007: nit - move this to test file as its being used as Oracle for new fn

pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -125,17 +126,18 @@ func (f *FirewallRuleProcessor) ComputeMapEntriesFromEndpointRules(firewallRules
...
-			cidrL4Info = checkAndDeriveL4InfoFromAnyMatchingCIDRs(string(firewallRule.IPCidr), nonHostCIDRs)
+			cidrL4Info = checkAndDeriveL4InfoFromAnyMatchingCIDRsTrie(string(firewallRule.IPCidr), containmentTrie, nonHostCIDRs)
 			if len(cidrL4Info) > 0 {
 				firewallRule.L4Info = append(firewallRule.L4Info, cidrL4Info...)
 			}
 		}
 		cidrsMap[string(firewallRule.IPCidr)] = firewallRule
 		if utils.IsNonHostCIDR(string(firewallRule.IPCidr)) {
+			_, alreadyInTrie := nonHostCIDRs[string(firewallRule.IPCidr)]
 			nonHostCIDRs[string(firewallRule.IPCidr)] = firewallRule
```
viveksb007, line 137: `nonHostCIDRs[string(firewallRule.IPCidr)] = firewallRule` line itself can go within the `if` condition below?

pkg/fwruleprocessor/cidr_trie_test.go
srikanthjg: was testing something and missed cleaning up

pkg/fwruleprocessor/fw_rule_processor.go
srikanthjg: Must stay outside — duplicate CIDRs need the merged L4Info written back to the map.

viveksb007 (approved):
lgtm, thanks for this cool improvement

PR #627, opened 2026-07-23
(npa) fix memory leak with stale/deleted pods in Pod Identifier to PE map

jaydeokar:
*Issue #, if available:*
An unbounded memory leak in the NPA controllers. podIdentifierToPolicyEndpointMap (MAP1) is populated cluster-wide (one entry per selector pod, on every node) and was effectively append-only. Under high pod churn with unique-named pods (e.g. Jobs with UUIDs), each churned pod created a permanent MAP1 entry that was never reclaimed → heap grows without bound.

Root cause - two independent defects

  1. Scope-blind stale detection — MAP1 inserts are cluster-wide, but MAP3 (networkPolicyToPodIdentifierMap, the diff basis for staleness) was built from local-only identifiers. Non-local identifiers could never be diffed out. #613 
  2. Purge gated on parentPEList — the purge loop was nested inside the live-PE loop and matched by exact PE name, so it no-oped on full NP delete and missed renamed/re-sliced PEs  #643 

*Description of changes:*

1. Make MAP3 cluster-wide — deriveTargetPods now returns a third value, allSelectedPodIdentifiers (every identifier the PE selects on any node). That set feeds both the stale diff and
  MAP3. Local sets still drive eBPF programming/cleanup (deliberately not widened).
2. Purge by parent NP name, ungated — moved the purge out of the parentPEList loop and switched to parent-NP-name matching, so it runs even when no live PEs remain (full delete) and
  catches renamed slices. New helpers deleteParentNPFromPodIdentifierMap / utils.DeleteParentNPFromPodIdentifierMap; drops only this NP's slices, keeps entries other NPs still reference.
3. cleanupPod baseline-restore — added an else branch to reset pod_state to baseline (DEFAULT_ALLOW, or DEFAULT_DENY in strict mode) when MAP1[id] is absent, fixing a pre-existing
  deselect bug where a deselected running pod kept stale enforcement.
4. Remove CPE controller's early return (CR blocker) — on len(parentCPEObjects)==0 the cluster controller returned early and skipped the stale purge, leaking on full CNP delete. Removed
  the return so it falls through, matching the PE path.
5. Gate baseline-restore on IsBPFContextRegistered (CR blocker / regression fix) — part 3 unconditionally wrote eBPF maps, which failed with "no bpf context registered" for already-torn-down pods and stalled policy programming for live pods in the same batch. New IsBPFContextRegistered method on the BpfClient interface skips the write when the pod's context is already gone.

### DIFF
--- a/controllers/clusterpolicyendpoints_controller.go
+++ b/controllers/clusterpolicyendpoints_controller.go
@@ -346,8 +346,9 @@ func (r *ClusterPolicyEndpointsReconciler) deriveClusterPolicyIngressAndEgressFi
 
 func (r *ClusterPolicyEndpointsReconciler) deriveTargetPodsForParentCNP(ctx context.Context, parentCNP, resourceName string) ([]npatypes.Pod, map[string]bool, []npatypes.Pod) {
 	var newTargetPods, podsToBeCleanedUp, currentPods []npatypes.Pod
-	var targetPodIdentifiers []string
-	podIdentifiers := make(map[string]bool)
+	var allSelectedPodIdentifiersSlice []string        // all identifiers this CNP selects (any node) - slice form
+	podIdentifiers := make(map[string]bool)            // node-local only - drives programming + cleanup diff
+	allSelectedPodIdentifiers := make(map[string]bool) // all identifiers this CNP selects (any node) - dedupe across CPEs
 
 	// Get current pods selected by the PE objects for this CNP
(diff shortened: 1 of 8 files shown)


yash97:
Can you please update PR description with problem and solution, for reference. Thanks

jaydeokar:
Working on retesting after fixing the merge conflicts

Issue #635, opened 2026-07-28
In strict mode, pods selected only by a ClusterNetworkPolicy do not get the implicit node-IP allow and kubelet probes are dropped


mharmer-canva:
**What happened**:

In `NETWORK_POLICY_ENFORCING_MODE=strict`, a pod whose only matching policy is a 
`ClusterNetworkPolicy` is enforced without the implicit node-IP allow that pods with
namespaced NetworkPolicies receive. Kubelet liveness/readiness probes are dropped, so
the pod crash-loops from its first probe. This contradicts the documented behaviour that
["node IP addresses are automatically allowed for kubelet health probes in strict mode"](https://repost.aws/articles/ARq5x53--OSPihwF3Y8C50wg/troubleshooting-kubernetes-network-policies-enforcing-policies-in-amazon-eks).

**Attach logs**

(can provide if necessary)

**What you expected to happen**:

Pods whose policy state comes from the cluster-policy path should receive the same
implicit node-IP allow as pods with namespaced policies.

**How to reproduce it (as minimally and precisely as possible)**:

1. EKS cluster with network policies enabled, `NETWORK_POLICY_ENFORCING_MODE=strict`,
   agent v1.3.7.
2. Apply a cluster-wide egress-only baseline policy:

```yaml
apiVersion: networking.k8s.aws/v1alpha1
kind: ClusterNetworkPolicy
metadata:
  name: allow-all-egress
spec:
  tier: Baseline
  priority: 50
  subject:
    namespaces: {}
  egress:
    - action: Accept
      to:
        - networks: [0.0.0.0/0, ::/0]
```

3. Deploy a workload with an httpGet liveness probe and no namespaced NetworkPolicy
   selecting it.
4. Every probe times out (`Liveness probe failed: ... context deadline exceeded`) and
   the pod crash-loops. Egress from the pod works throughout.
5. Apply any namespaced NetworkPolicy selecting the pod (even `policyTypes: [Ingress]`
   with no rules) — probes immediately recover. This is also the workaround.

**Anything else we need to know?**:

Root cause analysis:

`pkg/fwruleprocessor/fw_rule_processor.go`:

- `ComputeMapEntriesFromEndpointRules` (namespaced-policy path) unconditionally injects
  a node-IP allow entry — *"Traffic from the local node should always be allowed. Add
  NodeIP by default to map entries."*
- `ComputeClusterPolicyMapEntriesFromEndpointRules` (cluster-policy path) adds only the
[...]

PR #636, opened 2026-07-31
fix: add node-IP allow to ClusterNetworkPolicy ingress path

nakshah87:
In NETWORK_POLICY_ENFORCING_MODE=strict, pods whose only matching policy is a ClusterNetworkPolicy do not receive the implicit node-IP allow that pods with namespaced NetworkPolicies receive. This causes kubelet liveness/readiness probes to be dropped and the pod crash-loops.

Root cause:
The namespaced path (ComputeMapEntriesFromEndpointRules) unconditionally adds the node IP to the ingress firewall map, but the cluster policy path (ComputeClusterPolicyMapEntriesFromEndpointRules) was missing this entry.

Additionally, shouldSkipRule() discards any CNP rule whose CIDR matches the node IP exactly, making it impossible for users to work around this within a ClusterNetworkPolicy.

Fix:
Add the node-IP allow-all entry to ComputeClusterPolicyMapEntriesFromEndpointRules using ComputeTrieValueForCPE (the correct value format for cp_ingress_map, which includes a priority field — unlike the namespaced ingress_map which uses ComputeTrieValue with a 3-field format). 
The entry is written with Action=Accept at Priority=0 (highest), ensuring node-IP traffic is always allowed regardless of other CNP rules.

Testing:
- Reproduced on EKS cluster with v1.4.0 NPA in strict mode
- Applied egress-only ClusterNetworkPolicy + deployment with livenessProbe
- BEFORE fix: pod crash-looped every ~30s with: 'Liveness probe failed: context deadline exceeded' 
eBPF state: 
```
cp_ingress_map (Map 701): Empty ingress_map (Map 702): Empty 
ingress_pod_state_map: Key 0=2 (DEFAULT_DENY), Key 1=0 (POLICIES_APPLIED)
```
- AFTER fix (v1.4.0 + this patch): Pod stable 6+ minutes, 0 restarts, 1/1 Ready 
```
cp_ingress_map (Map 718): Contains node-IP allow entry (192.168.29.203/32) 
ingress_map (Map 719): Empty (expected - no namespaced NP) 
ingress_pod_state_map: Key 0=2 (DEFAULT_DENY), Key 1=0 (POLICIES_APPLIED)
```
 The probe traffic from node IP now matches in cp_ingress_map before falling through to the DEFAULT_DENY tier.

Unit tests added:
- TestFWRuleProcessor_ComputeClusterPolicyMapEntriesFromEndpointRules_NodeIPAlwaysPresent (4 sub-tests: empty rules, egress-only CNP, specific CIDR, node-IP CIDR)
- TestFWRuleProcessor_ComputeClusterPolicyMapEntriesFromEndpointRules_IPv6_NodeIPAlwaysPresent (2 sub-tests: empty rules, egress-only CNP)

*Issue #, if available:* https://github.com/aws/aws-network-policy-agent/issues/635

### DIFF
--- a/pkg/fwruleprocessor/fw_rule_processor.go
+++ b/pkg/fwruleprocessor/fw_rule_processor.go
@@ -284,6 +284,19 @@ func (f *FirewallRuleProcessor) ComputeClusterPolicyMapEntriesFromEndpointRules(
 	cidrL4Rules := make(map[string][]utils.L4Rule)
 	processedCIDRs := make([]string, 0)
 
+	// Traffic from the local node should always be allowed. Add NodeIP by default to map entries.
+	// This ensures kubelet liveness/readiness probes are never blocked by cluster network policies.
+	_, mapKey, _ := net.ParseCIDR(f.nodeIP + f.hostMask)
+	key := utils.ComputeTrieKey(*mapKey, f.enableIPv6)
+	nodeIPL4Rule := []utils.L4Rule{
+		{
+			Action:   "Accept",
+			Priority: 0, // Highest priority — node-IP allow must not be overridden
+		},
+	}
+	value := utils.ComputeTrieValueForCPE(nodeIPL4Rule)
+	firewallMap[string(key)] = value
+
 	// Step 1: Sort by CIDR length
 	f.sortByCIDRLength(firewallRules)
 
--- a/pkg/fwruleprocessor/fw_rule_processor_test.go
+++ b/pkg/fwruleprocessor/fw_rule_processor_test.go
@@ -754,3 +754,102 @@ func TestFWRuleProcessor_NonCanonicalCIDRCollision(t *testing.T) {
 		}
 	}
 }
+
+func TestFWRuleProcessor_ComputeClusterPolicyMapEntriesFromEndpointRules_NodeIPAlwaysPresent(t *testing.T) {
+	nodeIP := "10.1.1.1"
+	_, nodeIPCIDR, _ := net.ParseCIDR(nodeIP + "/32")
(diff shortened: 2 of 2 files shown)


pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -264,6 +264,19 @@ func (f *FirewallRuleProcessor) ComputeClusterPolicyMapEntriesFromEndpointRules(
...
 	processedCIDRs := make([]string, 0)
 
+	// Traffic from the local node should always be allowed. Add NodeIP by default to map entries.
+	// This ensures kubelet liveness/readiness probes are never blocked by cluster network policies.
+	_, mapKey, _ := net.ParseCIDR(f.nodeIP + f.hostMask)
+	key := utils.ComputeTrieKey(*mapKey, f.enableIPv6)
+	nodeIPL4Rule := []utils.L4Rule{
+		{
+			Action:   "Accept",
+			Priority: 0, // Highest priority — node-IP allow must not be overridden
```
cdirubbio, line 294: Wouldn't a deny at same priority still block this rule

nakshah87: CNP cannot express ingress CIDR rules today and even if it could in the future: LPM trie means /32 always wins over any broader CIDR.

nakshah87: For Namespaced NetworkPolicy and ApplicationNetworkPolicy, standard K8s NetworkPolicy is allow-list only and node-IP is injected unconditionally before any rules

cdirubbio (dismissed):

cdirubbio (approved):

PR #645, opened 2026-08-06
Fix stale eBPF entries when a pod is deselected from a PolicyEndpoint

Issacwww:
### Issue #, if available:  

When a pod stops being selected by a NetworkPolicy while the policy itself still exists — the `podSelector` is narrowed, or the pod loses the matching label — the nodeagent must clear that pod's eBPF maps. Today it does not: the pod stays at `POLICIES_APPLIED` with the previously programmed rules and traffic that no policy denies any more stays blocked until the pod is restarted.

Previously every dataplane update in `cleanupPod` sat behind `podIdentifierToPolicyEndpointMap.Load(podIdentifier)`, but `deriveTargetPodsForParentNP` had already dropped that identifier from the map. The guard was false, the whole body was skipped, and the function returned `nil` as if cleanup had succeeded.

### Description of changes:

The fix:

1. Adds the missing `else` ("no policies remain") branch to `cleanupPod`, which clears the ingress/egress maps with empty rule sets and resets `pod_state_map[POD_STATE_MAP_KEY]` to `DEFAULT_ALLOW`, or `DEFAULT_DENY` in strict mode.
2. Clearing with empty rules is safe because the guard is the absence of the *whole* map entry. `deriveTargetPodsForParentNP` iterates `parentPEList` and removes every PolicyEndpoint of the parent NP from a stale identifier, and `deletePolicyEndpointFromPodIdentifierMap` only drops the key once its list is empty — so a missing entry means no PolicyEndpoint selects the identifier at all. If another policy still applies, the entry survives and the existing re-derivation path runs unchanged.
3. Skips identifiers whose eBPF context is already gone instead of erroring: probes are detached once the last pod of an identifier leaves the node, so the map update would fail and make routine pod deletion requeue forever. This needs a read-only dataplane lookup, so `BpfClient` gains `HasBPFContext`.

No change to `deriveTargetPodsForParentNP`, `getPodListToBeCleanedUp`, `deriveStalePodIdentifiers`, the `deletePolicyEndpointFromPodIdentifierMap` call site, or `updateeBPFMaps`. This is the namespaced PolicyEndpoint counterpart of #605, which fixes the same defect on the ClusterPolicyEndpoint path.

## Test steps:

1. Deployed the PR image to the `aws-node` daemonset (`aws-eks-nodeagent` container) on a 2-node EKS cluster, VPC CNI v1.22.4, `NETWORK_POLICY_ENFORCING_MODE=standard`
2. Ran `make test-selector-narrowing CLUSTER_NAME=<cluster>` — three dataplane specs, each reading `egress_pod_state_map` through `aws-eks-na-cli` before and after the transition
3. `podSelector` narrowing: policy initially selects two clients, then is narrowed to one. Deselected client regains egress with no restart
4. Pod label removal: policy unchanged, the label is removed from the *live* pod. Client regains egress; the spec asserts UID, `creationTimestamp` and container restart counts are unchanged to prove no restart happened
5. Two policies, one pod (control case): the pod is dropped from one policy only. The surviving policy's rules must stay programmed; only after it is dropped from the second policy does the state fall back to `DEFAULT_ALLOW`

## Result: PASSED

```
Ran 3 of 9 Specs in 301.650 seconds
SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 6 Skipped
```

`egress_pod_state_map` key 0 — `0` = `POLICIES_APPLIED`, `1` = `DEFAULT_ALLOW`:

```
selectoropen@policy/egress_pod_state_map        key 0 state = 0   # policy selects it
selectoropen@policy/egress_pod_state_map        key 0 state = 1   # after narrowing

labelremovalclient@policy/egress_pod_state_map  key 0 state = 0   # label present
labelremovalclient@policy/egress_pod_state_map  key 0 state = 1   # after label removal

multipolicyclient@policy/egress_pod_state_map   key 0 state = 0   # both policies
multipolicyclient@policy/egress_pod_state_map   key 0 state = 0   # dropped from one - rules preserved
multipolicyclient@policy/egress_pod_state_map   key 0 state = 0
multipolicyclient@policy/egress_pod_state_map   key 0 state = 1   # dropped from both
```

The three consecutive `state = 0` readings for `multipolicyclient` are the evidence that clearing with empty rules does not over-clear: while one policy still selects the pod, its rules stay programmed and the state stays `POLICIES_APPLIED`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,16 @@ ENVTEST_K8S_VERSION = 1.25.0
 # Skip installing the latest managed addon while running cyclonus test
 SKIP_ADDON_INSTALLATION ?= "false"
 
+# Defaults for the integration test suite flags (see test/framework/options.go).
+KUBECONFIG ?= $(HOME)/.kube/config
+TEST_IMAGE_REGISTRY ?= registry.k8s.io
+AWS_REGION ?= us-west-2
+# IP family for the test suite. Derived from IP_FAMILY rather than defaulting it, so
+# targets that pass IP_FAMILY through to scripts keep their existing behaviour.
+TEST_IP_FAMILY = $(if $(IP_FAMILY),$(IP_FAMILY),IPv4)
+# Ginkgo focus regex matching only the PolicyEndpoint selector-narrowing specs.
+SELECTOR_NARROWING_FOCUS ?= PolicyEndpoint Selector Narrowing
+
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
 GOBIN=$(shell go env GOPATH)/bin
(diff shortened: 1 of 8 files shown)


Issacwww:
Closing this as change overlapped with https://github.com/aws/aws-network-policy-agent/pull/627

Issue #631, opened 2026-07-24
IPv6 except CIDRs in an IPv4 cluster produce malformed LPM keys and permanently abort egress map refreshes (including stale-entry deletion)


rogeriochaves:
**What happened**:

On an IPv4-only EKS cluster, a NetworkPolicy egress rule with an IPv6 ipBlock that carries `except` entries permanently breaks eBPF map reconciliation for every pod the policy selects. Every egress map refresh fails with:

```
{"level":"error","caller":"ebpf/bpf_client.go","msg":"Failed to update kernel map during bulk refresh for key \x40\x00\x00\x00\x01\x00\x00\x00: unable to update map: invalid argument"}
{"level":"error","caller":"ebpf/bpf_client.go","msg":"BPF map update failed unable to update map: invalid argument"}
{"level":"error","caller":"controllers/policyendpoints_controller.go","msg":"Map updates failed for podIdentifier <rs>@<ns>: egress map write: unable to update map: invalid argument"}
```

The failing key decodes to prefix length 64 (little endian `0x40 0x00 0x00 0x00`) with 4 bytes of data `0x01 0x00 0x00 0x00`, which is `100::/64` (one of our v6 except entries) encoded as an IPv4 LPM trie key. A prefix length of 64 on a 4-byte-data LPM trie is what the kernel rejects with EINVAL.

Because the refresh aborts on the first bad key, the pod's egress map silently freezes at whatever it contained when the probes were first attached. Endpoint churn (deploys of the destination workloads) leaves the map pointing at stale pod IPs, and the stale-key deletion phase never runs at all, so the map also retains allow entries for IPs that have since been recycled to unrelated pods. We observed 3042 aborted refreshes over 7 days in production before finding it; removing the IPv6 ipBlock from the policy made the very next reconcile succeed and purge 10 stale entries.

**Root cause** (present in v1.4.0 and still on main):

In `pkg/fwruleprocessor/fw_rule_processor.go`, `ComputeMapEntriesFromEndpointRules`:

1. The per-rule loop harvests `firewallRule.Except` into `exceptCidrs` BEFORE the `f.shouldSkipRule(...)` check that filters IPv6 CIDRs on IPv4 clusters. So the `::/0` rule itself is skipped correctly, but its v6 except entries survive.
2. The later `for exceptCidr := range exceptCidrs` loop adds each except as a deny entry into `cidrsMap` with no v4/v6 filtering of its own.
[...]

PR #647, opened 2026-08-10
fix(mismatch protocol in np): Fixing deadlock when cross protocol np rule is defined

zac-nixon:
*Issue #, if available:*
https://github.com/aws/aws-network-policy-agent/issues/631

*Description of changes:*

When v4 and v6 entries are available in the same NetworkPolicy, it can cause a deadlock on a v4 cluster when v6 rules are defined, because we were unconditionally adding the `except` block into the firewall wall rule generator. This PR ensures that v6 rules are never added in a v4 cluster, and vice-versa.

Testing locally in my own cluster.

### DIFF
--- a/pkg/fwruleprocessor/fw_rule_processor.go
+++ b/pkg/fwruleprocessor/fw_rule_processor.go
@@ -88,11 +88,6 @@ func (f *FirewallRuleProcessor) ComputeMapEntriesFromEndpointRules(firewallRules
 	sortFirewallRulesByPrefixLength(firewallRules, f.hostMask)
 
 	for _, firewallRule := range firewallRules {
-		// Keep track of except CIDRs to handle later
-		for _, exceptCidr := range firewallRule.Except {
-			exceptCidrs[string(exceptCidr)] = struct{}{}
-		}
-
 		var cidrL4Info []v1alpha1.Port
 
 		if !strings.Contains(string(firewallRule.IPCidr), "/") {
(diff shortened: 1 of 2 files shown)


pkg/fwruleprocessor/fw_rule_processor.go
```diff
@@ -115,6 +110,11 @@ func (f *FirewallRuleProcessor) ComputeMapEntriesFromEndpointRules(firewallRules
 			continue
 		}
 
+		// Track this rule's except CIDRs to handle later.
```
Pavani-Panakanti, line 114: I think this is also a valid policy and needs to be handled. Can we test it once ? Vice versa as well v6 cidr and v4 except
```
IPCidr: "10.0.0.0/8", Except: ["100::/64"]
```

zac-nixon: I don't think is a valid NP:
```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-egress-10-8
  namespace: default
spec:
  podSelector: {}
  policyTypes:
    - Egress
  egress:
    - to:
        - ipBlock:
            cidr: 10.0.0.0/8
            except:
              - "100::/64"
The NetworkPolicy "allow-egress-10-8" is invalid: spec.egress[0].to[0].ipBlock.except[0]: Invalid value: "100::/64": must be a strict subset of `cidr`
```

Basically, you can't mix a v4 and v6 cidr in the same `to` block.


Same thing in reverse case:
```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-egress-10-8
  namespace: default
spec:
  podSelector: {}
  policyTypes:
    - Egress
  egress:
    - to:
        - ipBlock:
            cidr: 100::/64
            except:
              - "10.0.0.0/8"
The NetworkPolicy "allow-egress-10-8" is invalid: spec.egress[0].to[0].ipBlock.except[0]: Invalid value: "10.0.0.0/8": must be a strict subset of `cidr`
```

Pavani-Panakanti: Great. Thanks for checking

Pavani-Panakanti (approved):

Issue #646, opened 2026-08-10
LoadBpfFile returns success with an incomplete map set; pod_state write then fails with EINVAL and the pod is isolated


tn20000:
**What happened**:

A cached eBPF map file descriptor in `policyEndpointeBPFContext` goes stale after a program reload, and every subsequent `ingress_pod_state_map` write for that `podIdentifier` fails with `EINVAL`. Because `UpdatePodStateEbpfMaps` swallows the error and the datapath treats a missing `pod_state` entry as `BPF_DROP`, the result is **permanent, total ingress blackholing of every pod sharing that `podIdentifier` on that node**, until the agent is restarted.

We hit this twice on one cluster. In the second occurrence, 7/7 pods of a churning pod set on the node were blackholed at birth over 57+ minutes. The agent never restarted (0 restarts), so this is not #514 — the trigger is an ordinary pod-driven program reload, not an `aws-node` recreate.

Log signature:

```
i Prog Load Succeeded for ingress, progFD: 45, pinpath: <podIdentifier>_handle_ingress
i Attempting to do an Ingress Attach with progFD: 45
i Pod has an Ingress hook attached. Update the corresponding map progFD: 45, mapName: ingress_pod_state_map, key: 0, value: {1}
e Ingress Pod State Map update failed: unable to update map: invalid argument
i Pod has an Egress hook attached. Update the corresponding map progFD: 49, mapName: egress_pod_state_map, key: 0, value: {1}
i Pod has an Ingress hook attached. Update the corresponding map progFD: 45, mapName: ingress_pod_state_map, key: 1, value: {1}
e Ingress Pod State Map update failed: unable to update map: invalid argument
```

Note egress (`progFD 49`) succeeds in the same call while ingress (`progFD 45`) fails — independent FDs go stale independently. Both keys 0 and 1 fail, so this is not a `max_entries` overflow.

**The affected pods had no NetworkPolicy selecting them, and that makes the impact worse rather than better.** `value: {1}` in those lines is `DEFAULT_ALLOW`, so this is `EnforceNpToPod` taking the "no active policies for this podIdentifier" branch and writing the *allow-everything* marker. The write failed, the entry was never created, and the datapath then dropped every packet — so a pod that no policy applies to was isolated precisely because its fail-open marker never landed. `AttacheBPFProbes` is called unconditionally before any policy check, so on a cluster with `enableNetworkPolicy: true` the exposure is every pod, not only policy-selected ones.

[...]

PR #648, opened 2026-08-10
fix(ebpf): reject a program load with an incomplete map set (pod_state EINVAL, #646)

tn20000:
## Summary

`LoadBpfFile` can return success with an incomplete direction-local map set. If that context is cached, a later pod-state update can index a zero-valued map and fail with `EINVAL`. This change makes the agent fail closed on unusable BPF state and preserves the currently attached datapath when a map write or attach operation fails.

## Changes

- Centralize the three required local maps for each TC direction (network policy, cluster policy, and pod state), and require a positive program FD plus nonzero map FDs before load, recovery, or map writes proceed.
- Reject incomplete recovery transactionally: hydrate all required shadow maps before publishing any recovered context, require every non-replaced direction, and propagate recovery failures to startup.
- Serialize attach, recovery publication, teardown, and map writes with stable per-pod locks. Existing TC filters are never detached on `EEXIST`; the attach error is returned while the existing filter and ownership state are preserved.
- Clean rejected successful loads using the SDK's separate loaded-map inventory, scope cleanup to the local program/maps, avoid double-close on retries, propagate failures through reconcilers and reattach, and cover the behavior with tests.

## Failure-handling contract

- A usable direction requires its program and all three direction-local maps. Missing or invalid state is an error, not an ordinary cache miss or a successful no-op.
- A pod-state write failure does not evict the context or shadow maps. The handles, attachment indexes, and ownership state remain available for a retry or diagnosis.
- Teardown removes context and shadow state only after both directions have been cleaned successfully; failed cleanup retains ownership state for retry.
- Rejected loads clean only resources attributable to that load. Shared global maps are not touched.

## Test Plan

- [x] Pre-merge Linux-focused package tests on `cfba3e9`:

  ```text
  $ docker run --rm -v /Users/fliang/.nora/worktrees/aws-network-policy-agent-648-minimal:/src -v /Users/fliang/go/pkg/mod:/go/pkg/mod:ro -w /src golang:1.26 go test ./pkg/ebpf ./pkg/utils ./controllers ./pkg/rpc
  ok   github.com/aws/aws-network-policy-agent/pkg/ebpf       0.045s
  ok   github.com/aws/aws-network-policy-agent/pkg/utils       0.006s
  ok   github.com/aws/aws-network-policy-agent/controllers     0.010s
  ok   github.com/aws/aws-network-policy-agent/pkg/rpc         0.012s
  ```

- [x] Merge conflict resolution against upstream `main` (`69fe6a5`) — `git diff --cached --check` clean; no unmerged index entries or conflict markers.
- [ ] Re-run the focused tests on merge head `f15c45e`; the local Docker daemon failed before container startup because a containerd layer was unavailable (`input/output error`).
- [ ] Race tests and integration tests require a working Linux environment; integration tests also require the Kubernetes test cluster/configuration.

## Deploy Plan

No relevant deploy plans.

### DIFF
--- a/pkg/ebpf/bpf_client.go
+++ b/pkg/ebpf/bpf_client.go
@@ -1339,14 +1339,14 @@ func (l *bpfClient) UpdatePodStateEbpfMaps(podIdentifier string, key int, state 
 		if !found {
 			sdkAPIErr.WithLabelValues("updateEbpfMap-egress-podstate-no-map").Inc()
 			egressErr = fmt.Errorf("map %s absent from bpf context for pod %s", utils.TC_EGRESS_POD_STATE_MAP, podIdentifier)
-			log().Errorf("Egress Map update failed: %v", egressErr)
+			log().Errorf("Egress Pod State Map update failed: %v", egressErr)
 		} else {
 			log().Infof("Pod has an Egress hook attached. Update the corresponding map progFD: %d, mapName: %s, key: %d, value: %d", egressProgFD, utils.TC_EGRESS_POD_STATE_MAP, keyval, podStateValue.state)
 			start := time.Now()
 			egressErr = mapToUpdate.CreateUpdateMapEntry(uintptr(unsafe.Pointer(&keyval)), uintptr(unsafe.Pointer(&podStateValue)), 0)
 			sdkAPILatency.WithLabelValues("updateEbpfMap-egress-podstate", fmt.Sprint(egressErr != nil)).Observe(msSince(start))
 			if egressErr != nil {
-				log().Errorf("Egress Map update failed: %v", egressErr)
+				log().Errorf("Egress Pod State Map update failed: %v", egressErr)
 				sdkAPIErr.WithLabelValues("updateEbpfMap-egress-podstate").Inc()
 				egressErr = fmt.Errorf("egress pod state map write: %w", egressErr)
 			}
--- a/pkg/ebpf/bpf_client.go
+++ b/pkg/ebpf/bpf_client.go
@@ -1271,15 +1271,23 @@ func (l *bpfClient) CreatePodStateEbpfEntryIfNotExists(podIdentifier string, key
 
 // evictBPFContextAfterPodStateFailure drops the cached BPFContext for podIdentifier after a
 // pod state map write has failed. The cached context can hold a program whose map set is
-// incomplete, and every write against that same context fails identically, so an in-process
-// retry cannot recover on its own. The PolicyEndpoint reconciler is the case that needs this:
-// it requeues with backoff and re-reads the cached context rather than reloading the program,
-// so without eviction it would retry the same unusable handles indefinitely. A pod state write
-// failure is already outage-grade (the datapath drops packets when the entry is missing), so
-// discard the context and let the next reconcile reload the program and re-derive its maps.
+// incomplete, and every write against that same context fails identically, so retrying
+// against the same cache entry can never recover. A pod state write failure is already
+// outage-grade -- the datapath drops packets while the entry is missing -- so discarding the
+// context is the cheaper risk: it forces the next attempt to program this podIdentifier to
+// reload the program and re-derive its maps instead of reusing the bad handles.
+//
+// Eviction is not itself a retry, and the callers differ in whether one follows. The
+// PolicyEndpoint and ClusterPolicyEndpoint reconcilers requeue with backoff and re-read the
+// context, so they recover; they are the case that needs eviction, because without it they
+// would retry the same unusable handles indefinitely. The RPC handler propagates the error to
+// the CNI, which fails the ADD and lets kubelet retry the sandbox. ReAttachEbpfProbes only
+// logs and continues, so nothing re-programs that podIdentifier until the next policy change
+// or agent restart; eviction still helps there by turning a poisoned cache entry into an
+// absent one, but it does not close that gap.
 func (l *bpfClient) evictBPFContextAfterPodStateFailure(podIdentifier string, op string) {
 	l.policyEndpointeBPFContext.Delete(podIdentifier)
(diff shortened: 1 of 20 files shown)


yash97:
Thanks for contribution, will review and test this pr, also looking for sdk fix mentioned. Will post update here in bundle

tn20000:
Hey thanks I pushed a new commit after upgrading to v1.4.1, seeing the error logs, and understood the errors a little bit more. Also added a new comment in #646 to explain it. Would appreciate your prompt help with the fix. 

AWS support case ID: 178639933100331

pkg/ebpf/bpf_client.go
```diff
@@ -1221,22 +1238,48 @@ func (l *bpfClient) CreatePodStateEbpfEntryIfNotExists(podIdentifier string, key
...
+// incomplete, and every write against that same context fails identically, so an in-process
+// retry cannot recover on its own. The PolicyEndpoint reconciler is the case that needs this:
+// it requeues with backoff and re-reads the cached context rather than reloading the program,
+// so without eviction it would retry the same unusable handles indefinitely. A pod state write
+// failure is already outage-grade (the datapath drops packets when the entry is missing), so
+// discard the context and let the next reconcile reload the program and re-derive its maps.
+func (l *bpfClient) evictBPFContextAfterPodStateFailure(podIdentifier string, op string) {
+	l.policyEndpointeBPFContext.Delete(podIdentifier)
+	log().Errorf("Evicted bpf context for podIdentifier %s after pod state %s failure; probes will be reloaded on the next reconcile", podIdentifier, op)
 }
```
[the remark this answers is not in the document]
tn20000: Good catch, and it was wrong for two of the three caller classes, not just one. Fixed in 572e01c.

`UpdatePodStateEbpfMaps` is reached from:

- the `PolicyEndpoint` / `ClusterPolicyEndpoint` reconcilers, which requeue with backoff and re-read the context, so a reconcile-driven reload really does follow;
- `pkg/rpc/rpc_handler.go`, which returns the error to the CNI, failing the ADD so kubelet retries the sandbox -- a retry, but not a reconcile;
- `ReAttachEbpfProbes`, which logs each failure and returns `nil`, so nothing re-programs that podIdentifier until the next policy change or agent restart.

The old wording named a mechanism that never runs in the last two, and it was worst exactly in the re-attach case -- the path where an operator reading the line most needs to know that no retry is coming.

The log now states what eviction did rather than who will act on it:

```
Evicted bpf context for podIdentifier %s after pod state %s failure; cached program handles are discarded, so the next attempt to program this podIdentifier will reload them
```

and the doc comment spells out the per-caller difference, including that eviction is still correct on the re-attach path (it turns a permanently poisoned cache entry into a merely absent one) but is not by itself a recovery.

On the second location you flagged: both sites call the one shared `evictBPFContextAfterPodStateFailure` helper, which holds the only copy of that log line, so this single change covers both.

The `ReAttachEbpfProbes` swallow itself is out of scope here -- that is a behavior change rather than a wording one, and it is the remaining permanent-isolation path from #646. I will send it as a separate PR.

Verified: `gofmt` clean, `GOOS=linux go build ./...` and `go vet ./pkg/ebpf/...` clean, and the full `pkg/ebpf` suite passes on linux/arm64 (29 tests, 40 subtests, 0 failures), including the `MissingMap` cases this PR added.

tn20000:
Also picked up the suppressed finding from the review body (`bpf_client.go:1342`, egress pod state log wording) in 34254ef.

It was worth taking rather than suppressing: both egress branches logged `Egress Map update failed`, which is the *same* string `UpdateEbpfMaps` logs for the policy map. So one log search spanned two different maps in two different functions with no way to tell them apart -- and the pod state map is the one whose absence drops the pod's traffic. Ingress was already specific, so this only brings egress in line; the policy map lines are unchanged.

pkg/ebpf/bpf_client.go
jaydeokar, line 1: While this will reject progs which does not have the required maps however it doesn't fix the root issue where a prog doesn't have all the maps. 
Can you share the NPA and SDK logs where you saw this issue on the node ?

tn20000:
```
  Test reproduction: vpc-cni v1.23.0-eksbuild.1,
  aws-network-policy-agent v1.4.1, ebpf-sdk v1.0.15
  (2026-08-14, UTC)

  04:35:01.340Z INFO  ebpf/bpf_client.go:793
    Prog Load Succeeded for ingress, maps: 2

  04:35:02.626Z ERROR rpc/rpc_handler.go:162
    Ingress Pod State Map update failed:
    unable to update map: invalid argument

  04:35:02.626Z ERROR rpc/rpc_grpc.pb.go:261
    Pod state map update failed while updating network policy state:
    ingress pod state map write: unable to update map: invalid argument
```
The failure occurred approximately 1.286 seconds after the loader reported maps: 2. The two error records are separate handler/RPC logging layers for one apparent map-update failure, not two independent failures. The complete unredacted bounded log window was provided to AWS Support in case 178639933100331.

pkg/ebpf/bpf_client.go
```diff
@@ -996,6 +996,23 @@ func (l *bpfClient) loadBPFProgram(fileName string, direction string,
 		return nil, -1, fmt.Errorf("program loaded but has no associated maps for pod %s direction %s progFD %d", podIdentifier, direction, progFD)
 	}
 
+	// A non-empty map set is not sufficient: the SDK has been observed returning a
```
viveksb007: this is just 1 of the map, we can add a correctness checker post the SDK loads the program and as part of that correctness check - we can check for all maps needed by ingress and egress program (could be separate set of maps). If correctness check fails, return err from here (we don't need to add guardrails all over the place).

also it would be great to reference these maps in single place instead of re-declaring it from maintenance perspective.

tn20000: Hey Vivek I updated the PR to address your comments and resolve merge conflicts. It's still up to your team to decide whichever parts you want to incorporate since you guys are the experts here

PR #649, opened 2026-08-11
handle errors with bpf state recovery

jaydeokar:
*Issue #, if available:*
Handle partial/failed BPF recovery from the updated SDK, and add an integration test verifying recovery across a same-binary agent restart.

*Description of changes:*
 Makes eBPF state recovery resilient to per-pin failures, fixes pin-path parsing for pod names containing dots, and adds integration coverage for recovery across a same-binary agent restart.

  - **`bpf_client.go`**: `recoverBPFState` / `ReAttachEbpfProbes` skip recovered entries with an unparseable pin path (empty identifier) instead of storing state under an empty key; log + metric on skip. Handle the SDK's partial recovery (non-nil error with usable state).
  - **Integration test** (`rollout_restart_recovery_test.go`): same-binary rollout restart reuses pinned programs (not reload), across standalone-dotted / Deployment / Job pods. Verified from host ground truth — CLI-parsed identifier vs name-derived, CLI prog IDs vs tc-attached (with hook inversion), pins + IDs preserved before/after, enforcement continuous.

### DIFF
--- a/pkg/clihelper/show.go
+++ b/pkg/clihelper/show.go
@@ -38,14 +38,30 @@ func formatLastSeen(lastSeen, now uint64) string {
 // Show - Displays all loaded AWS BPF Programs and their associated maps
 func Show() error {
 
-	bpfSDKclient := goelf.New(goelf.Config{NamespacedMaps: utils.NamespacedBPFMaps})
+	bpfSDKclient := goelf.New(goelf.Config{NamespacedMaps: utils.NamespacedBPFMaps, GlobalMaps: utils.GlobalBPFMaps, GlobalPinPrefix: utils.GLOBAL_PIN_PREFIX})
 	bpfState, err := bpfSDKclient.GetAllBpfProgramsAndMaps()
 	if err != nil {
 		return err
 	}
 
 	for pinPath, bpfEntry := range bpfState {
-		podIdentifier, direction := utils.GetPodIdentifierFromBPFPinPath(pinPath)
+		podIdentifier, progName, isGlobal := bpfSDKclient.GetProgIdentifierFromBPFPinPath(pinPath)
+		if isGlobal {
+			// Global programs are not associated with any pod, so skip them in this output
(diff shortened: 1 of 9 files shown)


jaydeokar:
Depends on - https://github.com/aws/aws-ebpf-sdk-go/pull/163

fabioaraujopt:
Reproduced the partial-recovery divergence this PR handles, on v1.2.7-eksbuild.1 / VPC CNI
v1.20.4-eksbuild.2 / SDK v1.0.17. Detail in a comment on aws/aws-ebpf-sdk-go#163; briefly:
recovery adopted 4 maps on a node holding 63, the agent then attached a duplicate ingress
program for a pod and bulk-refreshed the correct allow-list into the new map while the kernel
kept filtering the orphaned one. `awsnodeagent_aws_ebpfsdk_error_count{fn="RecoverAllBpfProgramAndMaps"} 1`,
and `controller_runtime_reconcile_total{result="success"} 230788` with zero reconcile errors —
so nothing indicated a problem. Net effect was a silent ingress black hole on a CoreDNS pod that
stayed Ready and in Endpoints; lookups went from ~10ms to 10–40s.

Strongly in favour of this landing. Two things it looks like it doesn't cover:

**A skipped pin leaves the old program attached.** In `ReAttachEbpfProbes`, `podIdentifier == ""`
skips reattach and increments `reattach-bad-pinpath`, but the TC filter from the previous agent
generation is still attached and still filtering. That's the state we ended up in — a program the
agent no longer tracks, enforcing a map it no longer updates. Detaching the orphan, or refusing to
attach a second program to an interface that already has one, would close that.

**Partial recovery may not be survivable.** `bpf_client.go` currently logs and moves on with
"We will overwrite and recreate the maps/programs" — in our case it created a duplicate instead,
and 230k reconciles made the divergence permanent rather than repairing it. Failing fast would
have been better than silently enforcing stale state.

PR #655, opened 2026-08-18
test: run cyclonus + integration suites on IPv4 and IPv6 in parallel

yash97:
## What

Make the `/test cyclonus` PR gate exercise **both IPv4 and IPv6** clusters in parallel, and broaden per-leg suite coverage.

Today the gate creates a single IPv4-only cluster and runs cyclonus + the policy suite. This adds a parallel IPv6 leg and runs the ebpf, clusternetworkpolicy, and strict suites too.

## Changes

**`.github/workflows/pr-cyclonus-tests.yaml`**
- Matrix `ip-family: [IPv4, IPv6]` with `fail-fast: false` — the two legs run in parallel and independently.
- One cluster per family via an eksctl **config file** (`kubernetesNetworkConfig.ipFamily` + `iam.withOIDC`, required for IPv6) with deterministic names `cyclonus-pr-<PR#>-<family>`.
- Single shared image build; each leg posts its own labeled comment: **`Cyclonus Test Results (IPv4/IPv6): ✅/❌`**.
- Per-PR `concurrency` group (queue, don't cancel) so overlapping comments can't collide on the cluster name; `timeout-minutes: 180` on the test job.

**`scripts/run-cyclonus-tests.sh`**
- Opt-in `RUN_EBPF_TESTS` and `RUN_CLUSTER_NETWORK_POLICY_TESTS` flags (default `false`, so `make run-cyclonus-test` / `update-image-and-test` are unaffected), plus existing `ENABLE_STRICT_MODE`.
- All suites run through a shared `run_ginkgo_suite` helper (dedups four near-identical ginkgo invocations). Strict runs **last** (it mutates `aws-node` to strict enforcement).

## Per-leg coverage

| Leg | Suites |
|---|---|
| **IPv4** | cyclonus + policy + ebpf + clusternetworkpolicy + strict |
| **IPv6** | cyclonus + policy + clusternetworkpolicy + strict (ebpf self-skips — AF_PACKET is IPv4-only) |

## Testing

- `bash -n` on the script and YAML parse on the workflow: clean.
- Suite dir → binary-name mapping verified (`ebpf.test`, `clusternetworkpolicy.test` match `make build-test-binaries`).
- Note: `issue_comment` workflows run the **default-branch** definition, so `/test cyclonus` can only exercise this new logic **after merge**. First post-merge run should confirm: IPv6 pods get IPv6 addresses, `aws-node` retains `ENABLE_IPv6=true` after the addon update, and the ClusterNetworkPolicy suite finds its CRD on the managed cluster.

### DIFF
--- a/.github/workflows/pr-cyclonus-tests.yaml
+++ b/.github/workflows/pr-cyclonus-tests.yaml
@@ -3,6 +3,10 @@ name: PR Cyclonus Tests
 on:
   issue_comment:
     types: [created]
+  # Manual trigger — run from any branch to exercise THAT branch's workflow
+  # definition (e.g. to validate this workflow on an upstream test branch
+  # before merging to main, without the /test cyclonus comment path).
+  workflow_dispatch:
 
 permissions:
   id-token: write
@@ -10,11 +14,11 @@ permissions:
   pull-requests: write
   issues: write
 
-# Serialize runs per PR so overlapping `/test cyclonus` comments cannot collide
-# on the deterministic cluster name (create-vs-delete race). Queue rather than
-# cancel — cancelling mid-run would leak clusters.
+# Serialize per PR (comment path) so overlapping `/test cyclonus` comments can't
+# collide on the deterministic cluster name; dispatch runs key on run_id so each
+# is independent. Queue rather than cancel — cancelling mid-run leaks clusters.
 concurrency:
-  group: cyclonus-tests-${{ github.event.issue.number }}
+  group: cyclonus-tests-${{ github.event.issue.number || github.run_id }}
   cancel-in-progress: false
 
 env:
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-cyclonus-tests.yaml
```diff
@@ -140,13 +110,20 @@ jobs:
...
-  run-cyclonus:
-    needs: [check-trigger, create-cluster, build-image]
-    if: always() && needs.create-cluster.result == 'success' && needs.build-image.result == 'success'
+  # One leg per IP family. Each leg creates its own cluster, runs the full
+  # suite set, and posts a family-labeled results comment. fail-fast: false so
+  # an IPv6 failure does not cancel the IPv4 leg (and vice versa).
+  cyclonus:
+    needs: [check-trigger, build-image]
     runs-on: ubuntu-latest
+    timeout-minutes: 180
```
Pavani-Panakanti: 180 min is not enough for cyclonus tests ?

.github/workflows/pr-cyclonus-tests.yaml
```diff
@@ -161,6 +138,41 @@ jobs:
...
+          apiVersion: eksctl.io/v1alpha5
+          kind: ClusterConfig
+          metadata:
+            name: ${CLUSTER_NAME}
+            region: ${REGION}
+            version: "${K8S_VERSION}"
+          iam:
+            withOIDC: true
+          kubernetesNetworkConfig:
+            ipFamily: ${IP_FAMILY}
```
Pavani-Panakanti, line 162: where are we taking care of patching np agent with enable ipv6 true flag in the PR ?

.github/workflows/pr-cyclonus-tests.yaml
```diff
@@ -79,45 +85,9 @@ jobs:
             await github.rest.issues.createComment({
               ...context.repo,
               issue_number: context.issue.number,
-              body: `✅ Cyclonus tests triggered by @${context.actor} for \`${requestedSHA}\`. [View run](${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId})`
+              body: `✅ Cyclonus tests (IPv4 + IPv6) triggered by @${context.actor} for \`${requestedSHA}\`. [View run](${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId})`
```
Pavani-Panakanti, line 88: Looks like the cyclonus workflow on this PR used main yaml and not the PR one - "✅ Cyclonus tests triggered by @yash97 for 5554aae99b92a58599cdaf32b0b4a80869b1493e. [View run](https://github.com/aws/aws-network-policy-agent/actions/runs/32195500688)"
Want to confirm this was tested in other ways ?

yash97: oh yeah i just tested cyclonus script with against ipv6.

.github/workflows/pr-cyclonus-tests.yaml
```diff
@@ -233,4 +258,4 @@ jobs:
           role-duration-seconds: 3600
 
       - name: Delete EKS cluster
-        run: eksctl delete cluster --name $CLUSTER_NAME --region $REGION
+        run: eksctl delete cluster --name $CLUSTER_NAME --region $REGION --disable-nodegroup-eviction || echo "cluster delete failed or cluster not found"
```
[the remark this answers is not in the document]
yash97: Fixed in `0a861e1` — the delete step now retries genuine failures (up to 3x) and only treats a real `ResourceNotFoundException` as already-deleted, so auth/throttling/dependency errors surface instead of being masked as success.

.github/workflows/pr-cyclonus-tests.yaml
yash97: Good catch — bumped to 330 min in `0a861e1`. cyclonus alone runs ~3h; a full leg (cyclonus + policy + ebpf + clusternetworkpolicy + strict + cluster setup) is ~4-4.5h, and 330 stays under the 6h credential / GitHub job cap.

.github/workflows/pr-cyclonus-tests.yaml
yash97: `--enable-ipv6=true` is set automatically by the managed vpc-cni addon on an IPv6 cluster, and `kubectl set image` only swaps the image (preserving the container args), so the flag is preserved — verified on a manual IPv6 run. One risk that run surfaced: eksctl 0.223.0 did not attach `AmazonEKS_CNI_IPv6_Policy` to the vpc-cni addon IRSA role, which failed the IPv6 nodegroup join. CI installs the latest eksctl (0.230.0), which handles addon IAM via Pod Identity and may not reproduce it — we will confirm on the first real IPv6 leg and add the policy to the cluster config if it does.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/aws/aws-network-policy-agent
 
-go 1.26.5
+go 1.26.6
```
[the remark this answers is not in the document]
yash97: go.mod now matches main (`go 1.26.6` via #658 merged into this branch), so there is no PR-specific go.mod change remaining. `.go-version` (`1.26`) resolves to the same 1.26.x, so the build and govulncheck jobs stay consistent.

Pavani-Panakanti (approved):

PR #657, opened 2026-08-19
fixed E2E metric calculations

### LINKED ISSUE #644: awsnodeagent_policy_programming_latency_seconds records inaccurate observations due to stale last-change-trigger-time annotation
**What happened**:

The end-to-end policy programming latency histograms `awsnodeagent_policy_programming_latency_seconds` and `awsnodeagent_cluster_policy_programming_latency_seconds` (added in #516) record large numbers of multi-minute observations on clusters whose dataplane is demonstrably healthy. Mechanism, confirmed against the code in both repos:

- The Network Policy Controller stamps `networking.k8s.aws/last-change-trigger-time` on a PolicyEndpoint in `setLastChangeTriggerTime()` (`pkg/policyendpoints/manager.go`, aws/amazon-network-policy-controller-k8s#213) only when processing a genuine change — but PolicyEndpoints are also rewritten at other times (background rewrites, mass rewrites during node churn) without the annotation being refreshed.
- The agent's `observePolicyProgrammingLatency()` (`controllers/policyendpoints_controller.go`, #516) computes `time.Since(triggerTime)` and calls `Observe()` on every PE reconcile. The only guard is trackerStartTime (annotations predating agent start are skipped); there is no per-PE comparison against the previously-seen annotation value. Since every agent watches every PE, each rewrite of a PE carrying an old annotation produces one observation of (now − old annotation) per agent, landing in the top (≥300s / +Inf) buckets.

An observation should be emitted only when a PE write actually carries a new trigger annotation: the agent should compare the annotation to the previously-seen value for that PE and skip unchanged ones. And/or the controller should refresh or clear the annotation on non-trigger rewrites so it never accompanies a write it did not trigger. kube-proxy's network_programming_duration_seconds handling of EndpointsLastChangeTriggerTime — which is cited as the pattern being followed — is the prior art: it only measures when the trigger time changes.

**How to reproduce it (as minimally and precisely as possible)**:

1. EKS Kubernetes 1.35, VPC CNI v1.22.3 with network policy enabled (aws-network-policy-agent v1.3.7, EKS-managed network policy controller).
2. Create a NetworkPolicy whose selector membership rarely changes (e.g. a static ingress policy for a stable Deployment).
3. Wait for, or induce, PolicyEndpoint rewrites that involve no selector-membership change (node churn / controller rewrite cycle).
4. Observe resourceVersion advances while the annotation stays fixed:
```
kubectl get policyendpoint <pe> -o jsonpath='{.metadata.resourceVersion} {.metadata.annotations.networking\.k8s\.aws/last-change-trigger-time}'
```
5. Scrape awsnodeagent_policy_programming_latency_seconds on any node: each such rewrite adds one observation of (now − annotation) per agent.

**Environment**:
Kubernetes version (use kubectl version): 1.35
CNI Version: 1.22.3
Network Policy Agent Version: 1.3.7

viveksb007:
*Issue #, if available:*
Fixes #644 

*Description of changes:*

Introduces a consume-once pattern:
  each reconciler keeps a `sync.Map` of the last consumed annotation value per
  PE (`types.NamespacedName`) / CPE (name), and emits an observation only when
  the annotation value changes. The trigger time is consumed *before* the
  suppression checks (programming failure, negative latency), so a suppressed
  observation can never be re-emitted inflated by a later resync. Tracking
  entries are removed when the PE/CPE is deleted. Memory cost is ~190 bytes per
  tracked PE, bounded by live PE count.

### Testing

  - Unit tests (`controllers/programming_latency_test.go`, run with `-race`):
    dedup, restart guard, consume-on-failure, clock-skew skip, cleanup reset,
    CPE parity.
  - Verified on a live EKS cluster (K8s 1.35, 2 nodes): metadata-only PE/CPE
    writes now produce zero observations (previously one ~12h observation per
    write per agent, landing in +Inf); genuine NetworkPolicy /
    ClusterNetworkPolicy changes are still observed exactly once per agent
    (~7 ms); agent restart and PE/CPE deletion emit nothing.

Map memory shouldn't increase much as data stored is PE key and timestamp value

  | PE count | map memory |
  |---|---|
  | 100 | ~19 KB |
  | 1,000 | ~190 KB |
  | 10,000 | ~1.9 MB |
  | 100,000 (extreme) | ~19 MB |

### DIFF
--- a/controllers/clusterpolicyendpoints_controller.go
+++ b/controllers/clusterpolicyendpoints_controller.go
@@ -82,6 +82,10 @@ type ClusterPolicyEndpointsReconciler struct {
 	// stale ClusterPolicyEndpoint annotations and avoid emitting artificially high
 	// E2E latency on restart
 	trackerStartTime time.Time
+	// lastObservedTriggerTimes maps CPE name (cluster-scoped, so name alone is
+	// unique) to the last last-change-trigger-time annotation value consumed,
+	// so each trigger time is observed at most once
+	lastObservedTriggerTimes sync.Map
 
 	// Maps pod Identifier to list of ClusterPolicyEndpoint resources
 	podIdentifierToClusterPolicyEndpointMap      sync.Map
@@ -124,6 +128,8 @@ func (r *ClusterPolicyEndpointsReconciler) reconcile(ctx context.Context, req ct
 
 func (r *ClusterPolicyEndpointsReconciler) cleanUpClusterPolicyEndpoint(ctx context.Context, req ctrl.Request) error {
 	log().Infof("Clean Up ClusterPolicyEndpoint resources for name: %s", req.Name)
+
+	r.lastObservedTriggerTimes.Delete(req.Name)
 
 	parentCNP := utils.GetParentNPNameFromPEName(req.Name)
 	resourceName := req.Name
@@ -173,6 +179,7 @@ func (r *ClusterPolicyEndpointsReconciler) reconcileClusterPolicyEndpoint(ctx co
 		return err
 	}
 
+	programmingSucceeded := true
 	for podIdentifier := range targetPodIdentifiers {
 		ingressRules, egressRules, err := r.deriveClusterPolicyIngressAndEgressFirewallRules(ctx, podIdentifier, ClusterPolicyEndpoint.Name, false)
 		if err != nil {
@@ -183,18 +190,20 @@ func (r *ClusterPolicyEndpointsReconciler) reconcileClusterPolicyEndpoint(ctx co
 		err = r.configureClusterPolicyBPFProbes(podIdentifier, targetPods, ingressRules, egressRules)
 		if err != nil {
 			log().Errorf("Error configuring Cluster Policy eBPF Probes %v", err)
+			programmingSucceeded = false
 		}
 	}
 
-	r.observeClusterPolicyProgrammingLatency(ClusterPolicyEndpoint)
+	r.observeClusterPolicyProgrammingLatency(ClusterPolicyEndpoint, programmingSucceeded)
 
 	return nil
 }
 
-// observeClusterPolicyProgrammingLatency reads the last-change-trigger-time annotation
-// from the ClusterPolicyEndpoint and emits the E2E latency histogram if the timestamp
-// is newer than this agent's start time.
-func (r *ClusterPolicyEndpointsReconciler) observeClusterPolicyProgrammingLatency(cpe *policyk8sawsv1.ClusterPolicyEndpoint) {
+// observeClusterPolicyProgrammingLatency emits the E2E latency histogram from
+// the last-change-trigger-time annotation. Each annotation value is consumed
+// at most once per CPE, even when the observation is suppressed (programming
+// failure, clock skew), so stale rewrites/resyncs/relists never re-observe it.
+func (r *ClusterPolicyEndpointsReconciler) observeClusterPolicyProgrammingLatency(cpe *policyk8sawsv1.ClusterPolicyEndpoint, programmingSucceeded bool) {
 	if cpe.Annotations == nil {
 		return
 	}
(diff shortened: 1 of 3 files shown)


controllers/clusterpolicyendpoints_controller.go
```diff
@@ -82,6 +82,10 @@ type ClusterPolicyEndpointsReconciler struct {
 	// stale ClusterPolicyEndpoint annotations and avoid emitting artificially high
 	// E2E latency on restart
 	trackerStartTime time.Time
+	// lastObservedTriggerTimes maps CPE name (cluster-scoped, so name alone is
+	// unique) to the last last-change-trigger-time annotation value consumed,
+	// so each trigger time is observed at most once
+	lastObservedTriggerTimes sync.Map
```
srikanthjg, line 88: we dont have concurrency here as MaxConcurrentReconciles though is set to 3; we dont use it; defaults to one. You can just go with a simple map? This map is not used in the CNI path.

controllers/policyendpoints_controller.go
```diff
@@ -126,6 +126,10 @@ type PolicyEndpointsReconciler struct {
 	// stale PolicyEndpoint annotations and avoid emitting artificially high
 	// E2E latency on restart
 	trackerStartTime time.Time
+	// lastObservedTriggerTimes maps a PE's types.NamespacedName to the last
+	// last-change-trigger-time annotation value consumed, so each trigger
+	// time is observed at most once
+	lastObservedTriggerTimes sync.Map
```
srikanthjg, line 132: same comment as CPE, can you check if it is needed in the CNI path? other wise we can simplify it.

controllers/clusterpolicyendpoints_controller.go
viveksb007: yes but better to use concurrent safe map, just in case anyone decides to increase the MaxConcurrentReconciles. At that point, it won't be feasible to audit codebase for places which are not concurrent safe.
lmk if you see any concern in using this map

controllers/policyendpoints_controller.go
viveksb007: same as above. yeah this is controller scoped only, not used via CNI path

controllers/clusterpolicyendpoints_controller.go
srikanthjg: it is better to avoid it when we dont need it. sync map adds additional unnecessary overhead of both time and memory. you can add a unit test to verify that.

viveksb007: i think this map time and memory is negligible in comparison to the work being done as part of controller reconcile as this is 1 part of the reconciliation flow.

how would you avoid/catch if someone does increase concurrent reconcile which will result in weird behavior? 
if we can catch that somehow during build/CI, then it makes sense to convert it into normal map otherwise i am thinking its better to keep it concurrent safe.

controllers/policyendpoints_controller.go
```diff
@@ -316,7 +323,23 @@ func (r *PolicyEndpointsReconciler) observePolicyProgrammingLatency(pe *policyk8
...
+	}
+	// Consume before the suppression checks below: a suppressed observation
+	// must never be re-emitted (inflated) by a later resync of the same annotation.
+	r.lastObservedTriggerTimes.Store(peKey, triggerTimeStr)
+	if !programmingSucceeded {
+		log().Debugf("Skipping E2E policy programming latency for PE %s/%s: programming failed", pe.Namespace, pe.Name)
+		return
+	}
 	latency := time.Since(triggerTime).Seconds()
+	if latency < 0 {
```
srikanthjg, line 338: good catch.

controllers/policyendpoints_controller.go
```diff
@@ -316,7 +323,23 @@ func (r *PolicyEndpointsReconciler) observePolicyProgrammingLatency(pe *policyk8
...
 		return
 	}
+	peKey := types.NamespacedName{Namespace: pe.Namespace, Name: pe.Name}
+	if prev, ok := r.lastObservedTriggerTimes.Load(peKey); ok && prev.(string) == triggerTimeStr {
+		return
+	}
+	// Consume before the suppression checks below: a suppressed observation
+	// must never be re-emitted (inflated) by a later resync of the same annotation.
+	r.lastObservedTriggerTimes.Store(peKey, triggerTimeStr)
+	if !programmingSucceeded {
```
srikanthjg, line 333: If programming fails transiently and the requeue(not succeeding in the first attempt) sees the same annotation value, 
the dedup skips it — is that observation permanently lost even if the retry succeeds - on the requeued attempt?

controllers/clusterpolicyendpoints_controller.go
srikanthjg: I do see a gain even without converting to map and this is more of a nitpick. Worth adding a comment on why the choice was taken for future readers both here and for MaxConcurrentReconciles so that we dont miss read it again.

controllers/policyendpoints_controller.go
viveksb007: > If programming fails transiently and the requeue(not succeeding in the first attempt) sees the same annotation value

don't think this will happen in current case as we are observing at the end. if some err happens annotation value will not be consumed. check - https://github.com/aws/aws-network-policy-agent/blob/faeb75ba495e4f32b66d0fd0d07063ee8eb270a8/controllers/clusterpolicyendpoints_controller.go#L161

so for cases where reconciliation is failing, E2E metric will be measured if that reconciliation eventually succeeds. Those failure cases should be visible by other metrics like reconcile_failed, requeue, etc (https://book.kubebuilder.io/reference/metrics-reference has detailed list)

srikanthjg: Thanks for the clarification

srikanthjg (approved):

PR #662, opened 2026-08-25
ci: Nightly kernel tests across supported EKS AL2023 kernel lines

yash97:
*Issue #, if available:* N/A

## What this PR does

Adds a nightly workflow (**Nightly AL2023 Kernel Tests**, 07:00 UTC + manual dispatch) that runs the Network Policy Agent's kernel-sensitive test suites against **every kernel line shipped by currently supported EKS AL2023 AMIs**, on both IPv4 and IPv6 clusters. Today the agent is only exercised on whatever kernel the latest default AMI happens to have; kernel-specific regressions (BPF verifier rejections, datapath behavior changes) on the other supported lines go unnoticed until customers hit them.

## What we test

Each matrix leg creates an EKS cluster whose node group is **pinned to the discovered AMI**, then:

1. **AMI + kernel verification** — asserts every node runs the expected AMI ID and the *exact* expected kernel package (`.status.nodeInfo.kernelVersion`), so results are attributable to a specific kernel.
2. **Agent install + rollout** (`scripts/cyclonus-cluster.sh install`) — enables network policy on the vpc-cni addon and rolls the nightly-built node agent image. This is the primary kernel gate: the BPF verifier accepts or rejects the agent's TC programs here.
3. **`policy.test`** — network policy enforcement through the agent's eBPF datapath.
4. **`strict.test`** — strict enforcement mode (runs last; mutates aws-node).

Both suites behave identically on IPv4 and IPv6, so every kernel line gets uniform dual-stack coverage (new `scripts/run-kernel-smoke-tests.sh`).

## How the matrix is discovered

`scripts/smoke/eks-kernel-matrix/run.sh` (read-only) queries `eks describe-cluster-versions` for versions in standard/extended support, resolves each latest AL2023 x86_64 standard AMI from public SSM parameters, parses the kernel package for the `AL2023_x86_64_STANDARD` column from the matching `awslabs/amazon-eks-ami` release notes, and selects the **lowest supported Kubernetes version per unique kernel line**. The matrix is self-maintaining — new/retired kernel lines adjust it without CI edits. Current result:

| Kernel | K8s | AMI kernel package |
|---|---|---|
| 6.1 | 1.31 | 6.1.180-225.360.amzn2023 |
| 6.12 | 1.33 | 6.12.100-125.179.amzn2023 |
| 6.18 | 1.36 | 6.18.41-94.142.amzn2023 |

`scripts/smoke/eks-kernel-matrix/test.sh` validates discovery offline (support filtering, column-aware parsing when AMI variants ship different kernels, per-line dedup) with mocked `aws`/`curl` — no credentials needed.

## Refactor: shared cluster lifecycle

`scripts/cyclonus-cluster.sh` extracts the EKS cluster lifecycle (`create` / `verify` / `install` / `delete` with retry that only trusts `ResourceNotFoundException`) previously inlined in the PR Cyclonus workflow. Both the PR bot and the nightly now use it, removing ~150 duplicated workflow lines. PR-bot behavior is unchanged; its cleanup job now checks out the validated PR SHA so the script exists before this PR merges.

## Cost / operational notes

- 6 clusters per night (3 kernels × 2 IP families), 3× t3.large each, ~1h per leg, deleted by independent retrying cleanup jobs (`fail-fast: false`, non-cancelling concurrency).

## Testing done

- `actionlint` on both workflows; `bash -n` on all scripts
- Offline discovery test (`test.sh`) incl. variant-kernel and unsupported-version cases
- Live discovery against us-west-2 (table above)
- `eksctl create cluster --dry-run` for pinned/unpinned IPv4 + IPv6 configs
- Mocked end-to-end runs of `run-kernel-smoke-tests.sh` (suite order, strict-mode mutation, failure propagation) and `cyclonus-cluster.sh` (create/verify/install/delete paths)
- `make build test`


scripts/smoke/eks-kernel-matrix/run.sh
```diff
@@ -0,0 +1,88 @@
...
+        index($0, "<summary><b>Kubernetes " target "</b></summary>") { section=1; next }
+        section && /<summary><b>Kubernetes / { exit }
+        section && /<th>Package<\/th>/ { header=1; column=0; next }
+        header && /<\/tr>/ { header=0; next }
+        header && /<th>/ {
+            column++
+            if (value($0) == "AL2023_x86_64_STANDARD") standard=column
+            next
+        }
+        section && standard && /<td>kernel[^<]*<\/td>/ { row=1; column=0; next }
```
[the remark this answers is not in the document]
yash97: The header index does not include the Package cell: the `<th>Package</th>` line matches the section trigger (`header=1; column=0; next`) and is consumed before any increment, so `column` counts only the data columns (NVIDIA=1, NEURON=2, STANDARD=3, ...). The data-row counter is symmetric — the `<td>kernel...</td>` label cell is consumed by the row trigger — so both counters share the same numbering, and `column < standard && column+span >= standard` tests exactly whether a (possibly colspan'd) cell covers the STANDARD column.

Verified empirically against the live `v20260818` release notes (both the rendered page and the raw API body): all 6 supported versions parse to the correct `AL2023_x86_64_STANDARD` kernel — `6.1.180-225.360.amzn2023` for 1.31/1.32, `6.12.100-125.179.amzn2023` for 1.33–1.35, `6.18.41-94.142.amzn2023` for 1.36 — matching the matrix in the PR description. The offline fixture in `test.sh` also pins the variant case where NVIDIA/NEURON columns carry a different kernel than STANDARD (`9.9.9-wrong-variant` guard), which would catch exactly the offset this comment describes.

No parser change made. Note that as of 6d57aba discovery fetches the release body via the GitHub Releases API rather than the rendered page, and `describe-cluster-versions` now passes `--include-all` explicitly (see the review thread's other comments).

scripts/eks-test-cluster.sh
```diff
@@ -0,0 +1,115 @@
...
+NODES_CAPACITY="${NODES_CAPACITY:-3}"
+
+require() {
+    local name
+    for name in "$@"; do
+        [[ -n "${!name:-}" ]] || { echo "$name is required" >&2; exit 1; }
+    done
+}
+
+create_cluster() {
```
Pavani-Panakanti, line 15: why not reuse these https://github.com/aws/aws-network-policy-agent/blob/main/scripts/lib/cluster.sh#L25 ? we can add any new parameters as required

scripts/eks-test-cluster.sh
```diff
@@ -0,0 +1,115 @@
...
+    version: latest
+    configurationValues: '{"enableNetworkPolicy":"true"}'
+    resolveConflicts: overwrite
+EOF
+    eksctl update addon -f /tmp/eks-test-addon.yaml --force
+    kubectl set image daemonset/aws-node -n kube-system aws-eks-nodeagent="$NODE_AGENT_IMAGE"
+    kubectl rollout status daemonset/aws-node -n kube-system --timeout=300s
+}
+
+delete_cluster() {
```
Pavani-Panakanti, line 92: same why not reuse existing one - https://github.com/aws/aws-network-policy-agent/blob/main/scripts/lib/cluster.sh#L67

scripts/run-kernel-smoke-tests.sh
```diff
@@ -0,0 +1,56 @@
...
+GINKGO_TEST_BUILD_DIR="$DIR/../test/build"
+
+: "${IP_FAMILY:="IPv4"}"
+: "${REGION:="us-west-2"}"
+: "${TEST_IMAGE_REGISTRY:="registry.k8s.io"}"
+: "${KUBE_CONFIG_PATH:=$KUBECONFIG}"
+
+TEST_FAILED="false"
+
+run_ginkgo_suite() {
```
Pavani-Panakanti: This is a copy of https://github.com/aws/aws-network-policy-agent/blob/main/scripts/run-cyclonus-tests.sh#L49..can we reuse ?

scripts/run-kernel-smoke-tests.sh
```diff
@@ -0,0 +1,56 @@
...
+
+echo "Nodes for cluster: $CLUSTER_NAME"
+kubectl get nodes -owide
+
+echo "Making ginkgo test binaries"
+(cd $DIR/.. && make build-test-binaries)
+
+run_ginkgo_suite policy.test 15m
+
+echo "Enable network policy strict mode"
```
Pavani-Panakanti: this is duplicate of https://github.com/aws/aws-network-policy-agent/blob/main/scripts/run-cyclonus-tests.sh#L130..lets reuse

scripts/run-kernel-smoke-tests.sh
```diff
@@ -0,0 +1,56 @@
...
+        --cluster-kubeconfig="$KUBE_CONFIG_PATH" \
+        --cluster-name="$CLUSTER_NAME" \
+        --test-image-registry="$TEST_IMAGE_REGISTRY" \
+        --ip-family="$IP_FAMILY" || TEST_FAILED="true"
+}
+
+echo "Nodes for cluster: $CLUSTER_NAME"
+kubectl get nodes -owide
+
+echo "Making ginkgo test binaries"
```
Pavani-Panakanti, line 32: This compiles the Ginkgo suite binaries inside each matrix leg, so the same policy.test and strict.test are built six times per night on six separate runners. Can we build test binaries ones in build job and publish them instead ?

Pavani-Panakanti:
> actionlint on both workflows; bash -n on all scripts

Can we add lint check to these files in our CI tests ?

scripts/run-kernel-smoke-tests.sh
yash97: this is good to have, single compilation is not necessity.

scripts/eks-test-cluster.sh
yash97: Looked at reusing it, and the blocker is the lifecycle model rather than missing parameters. lib/cluster.sh create_cluster writes eks-cluster.yaml into the working directory, and its paired delete_cluster deletes by that file. That contract only works when create and delete run in the same job on the same runner, which is how run-tests.sh uses it. This workflow deliberately splits create / verify / install / delete across separate GitHub jobs (cleanup runs on a fresh runner even when the test leg dies), so the lifecycle has to be name-based. On top of that the nightly needs AMI pinning (ami: field per nodegroup) and autoModeConfig disabled, which the lib version does not model. Retrofitting all of that into create_cluster would change a function that run-tests.sh, e2e-conformance and performance all depend on, which felt like the wrong blast radius for this PR. Happy to converge them in a follow-up, likely by making run-tests.sh consume eks-test-cluster.sh rather than the other way around.

scripts/eks-test-cluster.sh
yash97: This one I want to push back on directly: reusing lib delete_cluster here would be a regression. That version is `eksctl delete cluster -f ./eks-cluster.yaml || echo Cluster Delete failed` - it needs the config file from create (which does not exist on the cleanup job's fresh runner), and it swallows deletion failures. Cleanup reliability is the main risk of this PR (6 clusters per night), so the delete here is name-based with 3 retries and only treats ResourceNotFoundException from describe-cluster as proof of deletion; anything else fails the job loudly. If anything I would like the lib version to adopt these semantics in a follow-up.

scripts/run-kernel-smoke-tests.sh
yash97: Good catch, done in 068638b: run_ginkgo_suite now lives in scripts/lib/ginkgo-suites.sh and both run-cyclonus-tests.sh and run-kernel-smoke-tests.sh source it.

scripts/run-kernel-smoke-tests.sh
yash97: Done in 068638b: the strict-mode enable sequence is now enable_strict_mode in scripts/lib/ginkgo-suites.sh, called from both scripts.

scripts/run-kernel-smoke-tests.sh
yash97: Small correction after a second pass: the helper now lives in the existing scripts/lib/tests.sh rather than a new lib file (dc631a6), which is closer to what you suggested. It also validates its required variables at call time now.

scripts/run-kernel-smoke-tests.sh
yash97: Same here: enable_strict_mode ended up in scripts/lib/tests.sh (dc631a6), next to the other test orchestration helpers.

yash97:
> Can we add lint check to these files in our CI tests ?

I looked at adding a shellcheck gate here, but the existing scripts don't pass it. A CI check would fail immediately on about 18 findings, nearly all in scripts this PR never touches (lib/cluster.sh, lib/network-policy.sh, lib/cleanup.sh and others). Bringing that shared code compliant is out of scope for this PR since run-tests.sh, e2e-conformance and performance all depend on it. The scripts added here were written to pass shellcheck. We can track it as github issue.

Pavani-Panakanti (approved):

PR #682, opened 2026-09-18
fix(ebpf): Re-evaluating established connections in strict mode involving CNPs

### LINKED ISSUE #650: Established connections are not re-evaluated in strict mode when cluster network policies are removed
# Problem Description

In strict mode, when a pod loses its last ClusterNetworkPolicy and state flips from policies_applied to default deny, already-established connections keep passing traffic indefinitely. New connections are correctly denied. The cause is GET_CT_VAL in the eBPF datapath, which does not handle the DEFAULT_DENY pod state that strict mode uses.

How re-evaluation works

Each pod identifier has a pod-state map with two entries: entry 0 for NetworkPolicy state (POLICIES_APPLIED, or DEFAULT_ALLOW in standard mode / DEFAULT_DENY in strict mode when no NetworkPolicy applies), and entry 1 for ClusterNetworkPolicy state (POLICIES_APPLIED or DEFAULT_ALLOW).

The datapath evaluates policy once per connection and caches the verdict in aws_conntrack_map, stamped with a byte from GET_CT_VAL(entry0, entry1). Later packets compare the stored byte to the current one. Equal means reuse the cached verdict; different means re-run evaluateFlow. The byte changing is the only trigger for re-evaluating an established connection.

What happens

GET_CT_VAL tests its first argument only against DEFAULT_ALLOW and POLICIES_APPLIED. DEFAULT_DENY matches no branch and falls through to the default value, 7.
```
  ┌───────────────┬──────────────────┬──────┐
  │    entry 0    │     entry 1      │ byte │
  ├───────────────┼──────────────────┼──────┤
  │ DEFAULT_ALLOW │ POLICIES_APPLIED │ 4    │
  ├───────────────┼──────────────────┼──────┤
  │ DEFAULT_ALLOW │ DEFAULT_ALLOW    │ 2    │
  ├───────────────┼──────────────────┼──────┤
  │ DEFAULT_DENY  │ POLICIES_APPLIED │ 7    │
  ├───────────────┼──────────────────┼──────┤
  │ DEFAULT_DENY  │ DEFAULT_ALLOW    │ 7    │
  └───────────────┴──────────────────┴──────┘
  ```
Deleting the CNP moves entry 1 from POLICIES_APPLIED to DEFAULT_ALLOW. In standard mode that is 4 to 2, so the byte changes and the connection is re-evaluated and dropped. In strict mode it is 7 to 7, so the datapath returns BPF_OK without calling evaluateFlow, and the connection keeps flowing under a policy that no longer exists. There is a diff in standard mode vs strict mode evaluation here

The same collapse hides three other tightenings, all requiring strict mode plus a CNP: removing the last NetworkPolicy, adding a NetworkPolicy, and removing both at once. Four combinations are affected.

  Test

  Same operation both runs (delete the CNP), same long-lived TCP connection pinging every 3s, same node and agent. The only difference is entry 0: strict mode leaves it at DEFAULT_DENY, while a NetworkPolicy on the pod holds it at POLICIES_APPLIED.
```
  strict, entry0=DEFAULT_DENY (byte 7 -> 7)
    14:35:07  CNP deleted
    14:37:46  seq 69 OK        69 OK / 0 FAIL, still flowing after 2m39s
              new connections DENIED throughout

  standard + NetworkPolicy, entry0=POLICIES_APPLIED (byte 7 -> 5)
    14:49:59  seq 5 OK
    14:50:01  CNP deleted
    14:50:07  seq 6 FAIL       connection dropped in ~6s
```

Fix: 

paramater ordering need to be fixed here 
```
__u8 ct_pod_state_val = GET_CT_VAL(clusterpolicy_pst->state, pst->state);
```

### PR DESCRIPTION
**Pull request by ishani-das**
Fixes #650

## Background context
Each packet on an established connection is packed into one byte (via `GET_CT_VAL`), and that is compared against the cached byte for the `conntrack_key` 5-tuple (src_ip, src_port, dest_ip, dest_port, protocol) in `aws_conntrack_map`. If it does not match the already cached byte for the type, it will reevaluate by calling `evaluateFlow()`.

## Problem
`GET_CT_VAL` only has explicit branches for `DEFAULT_ALLOW` and `POLICIES_APPLIED` in its first argument, so `DEFAULT_DENY` fell through to the same cached byte used for `POLICIES_APPLIED`+`POLICIES_APPLIED`. Four combinations of cases are affected by this: 1) removing the last ClusterNetworkPolicy, 2) removing the last NetworkPolicy, 3) adding a NetworkPolicy, and 4) removing both the NetworkPolicy and ClusterNetworkPolicy at once.

Example (combination 1): in strict mode, a pod with no `NetworkPolicy` sits at `DEFAULT_DENY`, so the cached conntrack byte never changed when its `ClusterNetworkPolicy` state changed, and `evaluateFlow` never re-ran on already-established (cached) connections, allowing them to pass despite being in strict mode.

## Fix
Swapped the order in which `GET_CT_VAL` parameter args are called in all four traffic control eBPF programs so `DEFAULT_DENY` only appears as the second argument, where the logic is handled correctly to cover all branches.

`GET_CT_VAL(pst->state, clusterpolicy_pst->state)` -> `GET_CT_VAL(clusterpolicy_pst->state, pst->state)`

Before:
| | entry0 = `pst->state` | entry1 = `clusterpolicy_pst->state` | Computed val |
| --- | --- | --- | --- |
| CNP present | `DEFAULT_DENY` | `POLICIES_APPLIED` | 7 |
| CNP deleted | `DEFAULT_DENY` | `DEFAULT_ALLOW` | 7 |

After:
| | entry0 = `clusterpolicy_pst->state` | entry1 = `pst->state` | Computed val |
| --- | --- | --- | --- |
| CNP present | `POLICIES_APPLIED` | `DEFAULT_DENY` | 6 |
| CNP deleted | `DEFAULT_ALLOW` | `DEFAULT_DENY` | 3 |

## Testing
1. Manually verified the exact byte-level change against a live cluster before and after the fix, and observed the expected change in value for combination 1. ✅

**Setup**: strict mode, no `NetworkPolicy` on the client pod (`pst->state`= `DEFAULT_DENY`, `clusterpolicy_pst->state` = `POLICIES_APPLIED`), one `ClusterNetworkPolicy` allowing client→server egress on TCP/8080, one persistent HTTP-over-nc connection sending a request every second.

| Before (byte = 7 before deletion) | After (byte = 6 before deletion) | 
| --- | --- | 
| 10:25:36  CNP deleted <br> 10:25:47  seq=48   Conntrack Val - 7 <br> 10:26:05  seq=66   Conntrack Val - 7 <br> 10:26:23  seq=84   Conntrack Val - 7 <br> 10:26:40  seq=101  Conntrack Val - 7 <br> 10:26:58  seq=119  Conntrack Val - 7 <br> 10:27:16  seq=137  Conntrack Val - 7 <br> 10:27:34  seq=154  Conntrack Val - 7 <br> 10:27:51  seq=172  Conntrack Val - 7      (byte never changed: 7 -> 7) | 11:01:21  CNP deleted <br> 11:01:31  seq=46 <br> 11:01:49  seq=46 <br> 11:02:06  seq=46 <br> 11:02:23  seq=46 <br> 11:02:41  seq=46 <br> 11:02:58  seq=46 <br> 11:03:16  seq=46 <br> 11:03:33  seq=46 <br> ... <br> 11:11:19  seq=46   (still same several minutes later) |

**Before**: Byte stuck at 7 the entire time → evaluateFlow() never re-runs → established connection never re-evaluated (and thus seq numbers increase). 

**After**: The computed value after deletion was `GET_CT_VAL(DEFAULT_ALLOW, DEFAULT_DENY) = 3`. Since 3 != 6, the fast path misses and `evaluateFlow() `re-runs, correctly denying the packet. Note that on deny, `evaluateFlow` determines it is the `DEFAULT_DENY` case, returns `BPF_DROP`, thus the 3 is never written, and the old cache entry (6) is deleted by calling`bpf_map_delete_elem`.

2. Added `strict_mode_reeval_test.go`, covering all four affected combinations (remove last CNP, remove last NetworkPolicy, add a NetworkPolicy, remove both at once). All specs passed (2 pre-existing, 4 new). ✅
```
Running Suite: Strict Mode Network Policy Test Suite
Will run 6 of 6 specs
••••••
Ran 6 of 6 Specs in 215.437 seconds
SUCCESS! -- 6 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestStrictModeNetworkPolicy (215.44s)
PASS
```

Note: When writing a test where `NetworkPolicy` is deleted, it was discovered that `DeleteNetworkPolicy` has a bug where successful `NetworkPolicy` deletions were being reported back to callers as a failure. This was fixed by moving the `IsNotFound` check inside the scope where the real error is visible (now also matching the existing pattern in `DeleteClusterNetworkPolicy`).

## Running strict mode integration tests locally
Prereq: K8 cluster with NPA already deployed and running with strict mode enabled (`NETWORK_POLICY_ENFORCING_MODE=strict` on the `aws-node` DaemonSet's `aws-eks-nodeagent` container). 

From repo root:
```
export KUBECONFIG=$HOME/.kube/config   # or wherever your kubeconfig lives
kubectl config use-context <cluster-context>

go test -tags netgo,ebpf,core -run TestStrictModeNetworkPolicy -v -timeout 20m \
  ./test/integration/strict/... \
  -args -cluster-kubeconfig=$HOME/.kube/config -cluster-name=<your-cluster-name> -aws-region=<your-region>
```

## Reference (byte vals from eBPF traffic control programs)
```
16: #define POLICIES_APPLIED 0
17: #define DEFAULT_ALLOW    1
18: #define DEFAULT_DENY     2
```
```
30: #define CT_VAL_DEFAULT_ALLOW 0
31: #define CT_VAL_DEFAULT_ALLOW_DEFAULT_ALLOW 2
32: #define CT_VAL_DEFAULT_ALLOW_DEFAULT_DENY 3
33: #define CT_VAL_DEFAULT_ALLOW_POLICIES_APPLIED 4
34: #define CT_VAL_POLICIES_APPLIED_DEFAULT_ALLOW 5
35: #define CT_VAL_POLICIES_APPLIED_DEFAULT_DENY 6
36: #define CT_VAL_POLICIES_APPLIED_POLICIES_APPLIED 7
```
```
38: #define GET_CT_VAL(a, b) \
39:    ((a) == DEFAULT_ALLOW    && (b) == DEFAULT_ALLOW    ? CT_VAL_DEFAULT_ALLOW_DEFAULT_ALLOW    : \
40:     (a) == DEFAULT_ALLOW    && (b) == DEFAULT_DENY     ? CT_VAL_DEFAULT_ALLOW_DEFAULT_DENY     : \
41:     (a) == DEFAULT_ALLOW    && (b) == POLICIES_APPLIED ? CT_VAL_DEFAULT_ALLOW_POLICIES_APPLIED  : \
42:     (a) == POLICIES_APPLIED && (b) == DEFAULT_ALLOW    ? CT_VAL_POLICIES_APPLIED_DEFAULT_ALLOW  : \
43:     (a) == POLICIES_APPLIED && (b) == DEFAULT_DENY     ? CT_VAL_POLICIES_APPLIED_DEFAULT_DENY   : \
44:     (a) == POLICIES_APPLIED && (b) == POLICIES_APPLIED ? CT_VAL_POLICIES_APPLIED_POLICIES_APPLIED : \
45:     CT_VAL_POLICIES_APPLIED_POLICIES_APPLIED)   // <-- catch-all
```

### CODE REVIEW

File: test/integration/strict/strict_mode_reeval_test.go [hunk 1]
```diff
@@ -0,0 +1,681 @@
...
+					"persistent connection should receive a response before policy removal")
+			})
+
+			By("removing the final ClusterNetworkPolicy", func() {
+				err := fw.ClusterNetworkPolicyManager.DeleteClusterNetworkPolicy(ctx, clientCNP)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			By("verifying the established connection is re-evaluated and blocked", func() {
+				responseCount, err := persistentHTTPResponseCount(clientName)
```
emilyhuaa: There's a race between DeleteClusterNetworkPolicy returning and the pst map catching up. If the snapshot lands before reconciliation, the next packet or two will still get through and Consistently fails. Can we Eventually wait for the count to stop moving first, then snapshot? Same in the other three scenarios.

